PR #474, opened 2021-11-15
Enabled Workflow Dispatch for Wiki + Set Wiki Sidebar to include tokens

ahmadabdalla:
- Added Workflow dispatch for wiki triggers
- Updated sidebar to include the Parameter File Tokens Design


.github/workflows/platform.wiki-sync.yml
```diff
@@ -5,6 +5,7 @@
 name: '.Platform: Sync Docs/Wiki'
 
 on:
+  workflow_dispatch:
```
eriqua, line 8: This would allow wiki to be updated also manually from a different branch than main. Are we sure we want this? I'd suggest to have that happening only on pushes to the main branch

ahmadabdalla: Good point. In ADO, we could control to allow the run when branch is main by using approvals and checks. Is there something similar in GitHub?

eriqua: I'm having a hard time understanding the need of the manual trigger for the wiki in the first place, let's have a sync on that.

MrMCake: Somehow GH managed to miss some files on the latest PR merge. At least that's how it looked like to me.

What we can do, and would be rather easy, is to add the same condition we have for a module-publish here. That is, only run the wiki job when the branch is main/master

eriqua: What I'm not convinced of is: even if we add that condition (run only if in main) if we open to workflow dispatch we open the chance to modify that condition in ones' branch and allow the wiki update from a branch different than main

ahmadabdalla: Closing this PR and we can discuss again if there is a need for this.

PR #405, opened 2021-11-04
Users/torivara/virtual network gateways update

torivara:
# Change

Virtual Network Gateway with ExpressRoute type will not work if Public IP is set as zonal and gateway is set as regional. This means that the default settings will fail in some cases because PIP is zonal and gateway is not. Only tested with Express Route Gateway.

The proposed change is an update to the default value of publicIpZones and a conditional setting of zones on the automatically created pip.

Maybe this module should also support already existing pip-resources?

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I'm sure there are no other open Pull Requests for the same update/change
- [ ] My corresponding pipelines / checks run clean and green without any errors or warnings
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (readme)
- [x] I did format my code

PR #503, opened 2021-11-19
Deployment of virtual gateway is failing, as the IP address has not set a zone, although the VPN GW is zone aware.

rahalan:
# Change

If no PIP zone is defined, but a zone aware gateway will be deployed, a zone is set as default

## Type of Change

## Checklist

- [x] I did format my code


arm/Microsoft.Network/virtualNetworkGateways/deploy.bicep
```diff
@@ -182,6 +184,9 @@ var activeActive_var = (virtualNetworkGatewayType == 'ExpressRoute') ? bool('fal
 // Public IP variables
 var gatewayPipName1 = (length(gatewayPipName) == 0) ? '${virtualNetworkGatewayName}-pip1' : gatewayPipName[0]
 var gatewayPipName2 = activeActive_var ? ((length(gatewayPipName) == 1) ? '${virtualNetworkGatewayName}-pip2' : gatewayPipName[1]) : ''
+var varPublicIpZones = (empty(publicIpZones)) ? [
```
MrMCake, line 187: Hola @rahalan , can you please cross reference this change with [this former PR](https://github.com/Azure/ResourceModules/pull/405) and discuss it with @torivara?

MrMCake: Sounds to me like, instead of assuming a default, we should provide a parameter in the parameter file instead. However, I did not look into the details.

rahalan: @MrMCake there is a already a parameter for zones, but it defaults to none (which is ok). But if you use a gateway type that needs a zone, I think it makes sense to just default to zone 1, otherwise the deployment will fail. Alternatively we can say, it should fail and just change our parameter file to not have this scenario

MrMCake: It still looks to me we're just undoing the exact change that was implemented in the previous PR as a fix. We're just 'hiding' it in a variable. Maybe we should take a look together.

torivara: Why not do a conditional value here, @rahalan? The zone should be set only if the gateway is zone aware. There are several use cases where no zone is necessary, like dev/test. This should be an option.

rahalan: @torivara you are correct. I'll remove it. But will update the parameter file, as it will not deploy (zone aware VPN sku is set, but no PIP zone provided).

torivara: The parameter file absolutely needs to be deployable, therefore the publicIpZones parameter must match the gatewaySku. I might have missed this earlier, when I updated the files.

MrMCake (approved):

arm/Microsoft.Network/virtualNetworkGateways/deploy.bicep
MrMCake: With the explicit parameter file we should be good.

PR #626, opened 2021-11-28
Fixed incorrect token

MrMCake:
Fixed incorrect token


eriqua (approved):
🆗

MariusStorhaug (approved):
Looks good, and aligned with a plan that we will be adding tokens a bit later when we agree on which tokens and what store.

PR #707, opened 2021-12-02
Create subnet child-resource

MrMCake:
# Change

- Create subnet child-resource
- Updated documentation
- Updated parameter file

## Type of Change


arm/Microsoft.Network/virtualNetworks/deploy.bicep
```diff
@@ -112,34 +112,71 @@ resource virtualNetwork 'Microsoft.Network/virtualNetworks@2021-05-01' = {
...
+    doNotVerifyRemoteGateways: contains(peering, 'doNotVerifyRemoteGateways') ? peering.doNotVerifyRemoteGateways : true
+    useRemoteGateways: contains(peering, 'useRemoteGateways') ? peering.useRemoteGateways : false
+  }
+  dependsOn: [
+    virtualNetwork_subnets
+  ]
+}]
+
+// Remote to local peering (reverse)
+module virtualNetwork_peering_remote 'virtualNetworkPeerings/deploy.bicep' = [for (peering, index) in virtualNetworkPeerings: if (contains(peering, 'reversePeeringEnabled') ? peering.reversePeeringEnabled == true : false) {
```
ahmadabdalla, line 164: I think `reverse` can be confusing as it is not part of our original documentation. Maybe the switch `enableRemotePeering` is more aligned? or something around that? Maybe its more of a construct than a module aligned to the doco?

arm/Microsoft.Network/virtualNetworks/.parameters/vnetPeering.parameters.json
```diff
@@ -123,17 +25,11 @@
...
-            ]
-        },
-        "roleAssignments": {
-            "value": [
-                {
-                    "roleDefinitionIdOrName": "Reader",
-                    "principalIds": [
-                        "<<deploymentSpId>>"
-                    ]
+                    "useRemoteGateways": false,
```
ahmadabdalla, line 28: This is a bit confusing using both `remote` and `reverse`. I feel this is evolving more into a construct rather than a module as there is custom logic. Happy to be course corrected

ahmadabdalla (changes requested):

arm/Microsoft.Network/virtualNetworks/subnets/deploy.bicep
```diff
@@ -0,0 +1,113 @@
...
+@description('Optional. Application gateway IP configurations of virtual network resource.')
+param applicationGatewayIpConfigurations array = []
+
+@description('Optional. Array of IpAllocation which reference this subnet')
+param ipAllocations array = []
+
+@description('Optional. An array of service endpoint policies.')
+param serviceEndpointPolicies array = []
+
+var formattedServiceEndpoints = [for serviceEndpoint in serviceEndpoints: {
```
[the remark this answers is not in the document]
MrMCake: The locations property is not supported in the UI and also from the docs it seems unclear what to make of it.
Aside from that, it makes the consumption of the module more comfortable as folks don't have to replicate the endpoints syntax in the parameter file

arm/Microsoft.Network/virtualNetworks/.parameters/parameters.json
```diff
@@ -19,66 +19,18 @@
...
-                        {
-                            "service": "Microsoft.Sql"
-                        },
-                        {
-                            "service": "Microsoft.Storage"
-                        },
-                        {
-                            "service": "Microsoft.KeyVault"
-                        }
+                        "Microsoft.EventHub",
```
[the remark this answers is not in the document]
MrMCake: See other comment

arm/Microsoft.Network/virtualNetworks/.parameters/parameters.json
```diff
@@ -19,66 +19,18 @@
                 {
```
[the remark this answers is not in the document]
MrMCake: added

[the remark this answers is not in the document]
MrMCake: and tested

arm/Microsoft.Network/virtualNetworks/deploy.bicep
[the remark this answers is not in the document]
MrMCake: Aside the fact that the parameter is default `false` and optional - shouldn't we be able to expect somebody who is using any module to understand for permissions they'd need to deploy a service? I mean - at what point do we start to replicate the official docs that we then have to maintain too? In this particular case I'd expect a user to know/understand that they'll require certain permissions if they want to also deploy the peering in the opposite direction.

MrMCake: Regarding the names I agree. However, regarding the logic itself I'd not for a few reasons
1. I was already asked why the feature isn't there (not the best defense, but still)
2. The 2-way peering is strongly correlated with the rest of the logic
3. It's an optional additional deployment that somebody can enable/disable and still algined to the documentation. The additional parameter I add here are not e.g. on the peering child-resource itself, but the parent. For similar reasons we e.g. need to use enable-flag for the VM extensions.
However, you're correct that it introduces 'additional' logic and hence is up for debate. I'd suggest to either bring it up in the channel or one of the next meetings

arm/Microsoft.Network/virtualNetworks/.parameters/vnetPeering.parameters.json
MrMCake: See other comment

arm/Microsoft.Network/virtualNetworks/deploy.bicep
MrMCake: Updated all references of `reverse` to `remote`.

arm/Microsoft.Network/virtualNetworks/subnets/deploy.bicep
```diff
@@ -0,0 +1,113 @@
...
+}
+
+@description('The resource group the virtual network peering was deployed into')
+output subnetResourceGroup string = resourceGroup().name
+
+@description('The name of the virtual network peering')
+output subnetName string = subnet.name
+
+@description('The resource ID of the virtual network peering')
+output subnetResourceId string = subnet.id
```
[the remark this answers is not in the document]
MrMCake: added

arm/Microsoft.Network/virtualNetworks/.parameters/parameters.json
```diff
@@ -89,28 +41,8 @@
...
-                },
-                {
-                    "name": "sxx-az-subnet-x-005",
+                    "name": "sxx-az-subnet-x-003",
                     "addressPrefix": "10.0.6.0/24",
-                    "networkSecurityGroupName": "",
-                    "routeTableName": "",
-                    "serviceEndpoints": [],
-                    "delegations": [],
                     "privateEndpointNetworkPolicies": "Disabled", // This property must be set to disabled for subnets that contain private endpoints. It confirms that NSGs rules will not apply to private endpoints (currently not supported). Default Value when not specified is "Enabled".
```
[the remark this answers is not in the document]
MrMCake: Moved comment with reference to `Considerations` section of subnet module ReadMe

arm/Microsoft.Network/virtualNetworks/.parameters/min.parameters.json
```diff
@@ -0,0 +1,22 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "name": {
+            "value": "sxx-az-vnet-min-001"
```
eriqua, line 6: I'd call it `sxx-az-vnet-x-002` for consistency with other modules

MrMCake: I thought the `x` is there to allow us to give it a a dedicated name :D Let's discuss naming next week :)

arm/Microsoft.Network/virtualNetworks/deploy.bicep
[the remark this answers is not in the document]
ahmadabdalla: fair callout and feel free to close. I remember this use case from a customer that had a service principal deploying the hub and establishing the local peering but was failing to create the reverse peering with the spokes as they resided in another management group and subscriptions, hence we had to explicitly give it permissions to do the remote peering from the doco reference.

arm/Microsoft.Network/virtualNetworks/subnets/deploy.bicep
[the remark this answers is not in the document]
ahmadabdalla: I've just tested this by hardcoding an array of locations:
```bicep
var formattedServiceEndpoints = [for serviceEndpoint in serviceEndpoints: {
  service: serviceEndpoint
  locations: [
    'eastus'
    'westeurope'
    'uaecentral'
    'brazilsoutheast'
  ]
}]
```
And monitored the route table on a test VM to see if it constrained the ranges to these locations only, but it did nothing except add 2 locations (northeurope, westeurope). In fact, looking at the exported template I saw this:
```json
 "locations": [
      "westeurope",
      "northeurope"
  ]
```
Which I think it only enables service endpoints to the main region and the region pair, but not others. So, I don't mind not having it and simplifying. No further comments from myside.

arm/Microsoft.Network/virtualNetworks/deploy.bicep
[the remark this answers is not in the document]
MrMCake: And you're right on that and I assume for many cases customers would e.g. have teams deploy the reverse/remote peering using that dedicated template. However, I'd appreciate to keep the functionality in for those cases where teams have the required permissions. Thanks for elaborting @ahmadabdalla

arm/Microsoft.Network/virtualNetworks/.parameters/min.parameters.json
MrMCake: Using parameter-`prefix` for `x`

arm/Microsoft.Network/virtualNetworks/subnets/readme.md
```diff
@@ -0,0 +1,73 @@
...
+
+The `privateEndpointNetworkPolicies` property must be set to disabled for subnets that contain private endpoints. It confirms that NSGs rules will not apply to private endpoints (currently not supported, [reference](https://docs.microsoft.com/en-us/azure/private-link/private-endpoint-overview#limitations)). Default Value when not specified is "Enabled".
+
+
+## Outputs
+
+| Output Name | Type | Description |
+| :-- | :-- | :-- |
+| `subnetAddressPrefix` | string | The address prefix for the subnet |
+| `subnetAddressPrefixes` | array | List of address prefixes for the subnet |
```
ahmadabdalla, line 66: @MrMCake , this one is still failing because it is optional and may not exist
```bash
The template output
     | 'subnetAddressPrefixes' is not valid: The language expression
     | property 'addressPrefixes' doesn't exist, available properties
     | are 'provisioningState, addressPrefix
```
We can just keep the first one as its a 'required' property and will always be available. The property may not exist hence its attempting to find it. Or maybe something using the `contains()` function and outputting `[]` if it doesn't exist?

ahmadabdalla (changes requested):
just a single comment on the failure bit for output from my end. Otherwise all good

arm/Microsoft.Network/virtualNetworks/subnets/readme.md
MrMCake: Added a condition

MrMCake: And successfully tested

ahmadabdalla (approved):
No further comments from my side. Approved 🚀

PR #714, opened 2021-12-04
Role Assignments: Principal Type Update + Uplift API Version for Module

ahmadabdalla:
# Change

- Updated to the latest API version for Role Assignments `2021-04-01-preview`. 

  > Wanted to use the latest referenced API from the [Role Assignments Doc](https://docs.microsoft.com/en-us/azure/templates/microsoft.authorization/2020-10-01-preview/roleassignments?tabs=bicep), but that is broken and is being tracked in this [issue](https://github.com/Azure/azure-rest-api-specs/issues/15604#issuecomment-896080064). Also allowing Bicep to support the latest API for intellisense is tracked [here](https://github.com/Azure/bicep/issues/784#issuecomment-896046402) so it should get updated in future Bicep releases as definitions are coupled with the Bicep version.

- Constrained the `principalType` property to the allowed values as per the documentation, with support for `''` as the default so that it is used when it is needed. This required the [any()](https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions-any) function to wrap around the `principalType` property values in the resource defintiion. This was present when using known API versions (i.e. `2020-08-01-preview`) to Bicep but won't complain with the latest API `2021-04-01-preview` as it doesn't exist yet in Bicep.

- Updated the readme

## Notes

- The readme file - Template references section - Role Assignment documentation, will NOT function with the latest API as it does not exist yet, hence I have reverted it to the latest 'working' version which is [2020-08-01-preview](https://docs.microsoft.com/en-us/azure/templates/microsoft.authorization/2020-08-01-preview/roleassignments?tabs=bicep). 
- Another option would be to swap the API in the resource to use `2020-08-01-preview` instead of `2021-04-01-preview` but I think we should be using the latest and all these issues referenced above will get resolved eventually.

## Testing

[Authorization: RoleAssignments](https://github.com/Azure/ResourceModules/actions/workflows/ms.authorization.roleassignments.yml)

## Type of Change

## Checklist

- [ ] I did format my code


arm/Microsoft.Authorization/roleAssignments/.bicep/nested_rbac_rg.bicep
```diff
@@ -316,13 +324,13 @@ var builtInRoleNames_var = {
...
 
-resource roleAssignment 'Microsoft.Authorization/roleAssignments@2020-04-01-preview' = {
+resource roleAssignment 'Microsoft.Authorization/roleAssignments@2021-04-01-preview' = {
   name: guid(subscriptionId, resourceGroupName, roleDefinitionId_var, principalId)
   properties: {
     roleDefinitionId: roleDefinitionId_var
     principalId: principalId
     description: !empty(description) ? description : null
-    principalType: !empty(principalType) ? principalType : null
+    principalType: any(!empty(principalType) ? principalType : null)
```
MrMCake, line 333: Are you sure the `any` is needed here? Linter does not light up in my case: 
image

ahmadabdalla: Have you switched the API to a currently known one by bicep? latest is `2020-08-01-preview`

image

I'm running on version `0.4.1.008`

MrMCake: Just checked your branch out

ahmadabdalla: Here is the message:
```
The property "principalType" expected a value of type "'ForeignGroup' | 'Group' | 'ServicePrincipal' | 'User' | null" but the provided value is of type "'' | 'Device' | 'ForeignGroup' | 'Group' | 'ServicePrincipal' | 'User' | null".bicep(BCP036)
```

I'd wrap the `principalType` as it is the one containing `''` as linter indicates that `null` is supported. Thoughts?
```bicep
principalType: !empty(principalType) ? any(principalType) : null
```

ahmadabdalla: @MrMCake it is now like this `any(principalType)`. Tested again against the dependency pipeline and the module parameters. All working

PR #746, opened 2021-12-06
Added Broken Link Check Action

ahmadabdalla:
# Change

- Added a new Broken Link Checker - [Lychee](https://github.com/lycheeverse/lychee)
  - Validates all URLs inside the repository
  - Runs on a Pull request against main for anything *md, also supports workflow_dispatch so it can be checked in a branch.
  - Created an exclusion file `.lycheeignore` and added a lot of the URLs that either don't resolve or has authorization to it.
- Added feature to the `Set-ModuleReadMe.ps1` utility to validate a generated template reference URL if it is broken or not
  - Checks the generated URL (with API version), if it works, then that is used. If not
  - It checks the same URL but removes the specific API version so we can use the latest published, if it works then that is used. If not
  - Checks if the resource is a child, if yes then use the Parent as reference using the latest API available. If not
  - Take the basic URL for [Bicep and Templates documentation](https://docs.microsoft.com/en-us/azure/templates/)
  - We will always aim to have a valid URL in there, hence all broken URLs are now fixed by using the action we introduced.
- Other:
  - Aligned all fake email addresses and URLs so that we can apply a consistent exclusion.
  - There were other paths inside the wiki that started with `.\` changed to `./` to align with the rest

Testing:
- This has been tested in the fork using a branch that attempts to merge to main with a broken URL, or no broken URLs. 

Example:

image

image

***Feel free to remove this sample text***
Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context.
List any dependencies that are required for this change.

## Type of Change

## Checklist

- [ ] I did format my code


utilities/tools/Set-ModuleReadMe.ps1
```diff
@@ -330,9 +330,29 @@ function Set-TemplateReferencesSection {
     $TextInfo = (Get-Culture).TextInfo
     foreach ($resourceType in $relevantResourceTypes) {
         $Type, $Resource = $resourceType.Type -split '/', 2
```
MrMCake, line 332: I know that was my code, but still: Can we rename that to `$provider, $resourceType = `?

utilities/tools/Set-ModuleReadMe.ps1
```diff
@@ -330,9 +330,29 @@ function Set-TemplateReferencesSection {
...
-        $sectionContent += ('- [{0}](https://docs.microsoft.com/en-us/azure/templates/{1}/{2}/{3})' -f $TextInfo.ToTitleCase($Resource), $Type, $resourceType.ApiVersion, $Resource)
+        $ResourceReferenceTitle = $TextInfo.ToTitleCase($Resource)
+        # Validate if Reference URL Is working
+        $TemplatesBaseUrl = 'https://docs.microsoft.com/en-us/azure/templates'
+        try {
+            $ResourceReferenceUrl = '{0}/{1}/{2}/{3}' -f $TemplatesBaseUrl, $Type, $resourceType.ApiVersion, $Resource
+            Invoke-WebRequest -Uri $ResourceReferenceUrl | Out-Null
```
MrMCake, line 338: Change to `$null = ` instead of `Out-Null`.

Is used to do that too until somebody told me it is less performant

ahmadabdalla: good one. Did read about it and you're right. Thanks for the tip. Updated :)

utilities/tools/Set-ModuleReadMe.ps1
ahmadabdalla: Updated to use `$ProviderNamespace` as its more aligned to what we see in the docs, and `$ResourceType`, had to update other things as well like the foreach and `SectionContent`, Please check.

MrMCake (approved):

PR #814, opened 2021-12-15
Updated the use dependency filter to work with just a resourceId

MrMCake:
Updated the use dependency filter to work with just a resourceId


eriqua (approved):
🆗

PR #895, opened 2022-01-17
Replaced CuaID implementation (to be discussed)

MrMCake:
# Change

- Replaced CuaID implementation with single-file solution
- Added repository setting for telemetry GUID
- Added implementation to optionally pass ID to template via pipeline (both GH & ADO)

Pipeline reference
| Platform | Pipeline |
| - | - |
| GitHub | [AnalysisServices: Servers](https://github.com/Azure/ResourceModules/actions/workflows/ms.analysisservices.servers.yml)
| Azure DevOps | [IaCS](https://dev.azure.com/servicescode/infra-as-code-source/_build/results?buildId=145337&view=results) |

## Type of Change


mblant (changes requested):
Can we change enableDefaultTelemetry to true by default?

MrMCake:
> Can we change enableDefaultTelemetry to true by default?

Hey @mblant, it is actually enabled by default via the `settings.json` config that is injected into each CI deployment. However, I now set it to `true` in the templates themselves.

From a consumer perpective, that means, they'd need to actively it in the `settings.json` when running the CI pipelines, and actively add a parameter to set the telemetry to false when building a solution with the templates.

MariusStorhaug (approved):
Lets just get it in.

PR #925, opened 2022-01-24
Updated ADO pipeline names & update GitHub workflow conditions to not skip over tests

MrMCake:
# Change

- Updated workflow conditions to not skip over tests
- Aligned Azure DevOps pipeline files with GitHub composite actions names

Pipeline reference
- Expected continous run when first pester task fails: https://github.com/Azure/ResourceModules/runs/4920105211?check_suite_focus=true (Note: The API tests ran now after all)
- Exected cancellation when job is cancled: https://github.com/Azure/ResourceModules/actions/runs/1739266888  (Note: This test is supposed to ensure cancellations still work)

## Type of Change


utilities/tools/AzureDevOps/src/devOpsPipelineTemplate.yml
```diff
@@ -41,12 +41,12 @@ stages:
   - stage: Validation
     displayName: Pester tests
     jobs:
-      - template: /.azuredevops/pipelineTemplates/module.jobs.validate.yml
+      - template: /.azuredevops/pipelineTemplates/jobs.validateModulePester.yml
```
eriqua, line 44: Is this template used or intended just as a reference?

MrMCake: It's a template for a tool that I wrote back in the day to generate the ADO pipelines from the GitHub counterpart. Thinking of it we may want to question whether we want to keep it or not. What do you think?

MrMCake: I guess it's more to maintain and now we have 'permanent' ADO pipelines

MrMCake: I'd almost be tempted to remove the logic as part of this PR

eriqua: If not used I'd opt for removing it

MrMCake: Removed

eriqua (approved):
🆗

Issue #462, opened 2021-11-11
Document how to update dependencies for compute images


rahalan:
If users use a different test environment, gallery images may not be found, as storage account name is generated automatically. That needs to be updated manually.

MrMCake:
@eriqua, is this issue still valid with the latest code base?

eriqua:
This has been solved by dependency pipeline update taking care of generating and backing up the vhd needed by the computer images module

PR #929, opened 2022-01-26
Update dependency pipeline with image template and vhd copy

eriqua:
# Change

Closes issue #462 

This change allows the creation of a custom image using Azure VM Image Builder in both the GH and ADO dependency pipelines. 
In particular, the additional jobs are:
- deploying the Azure Image Builder template
- triggering the image build and waiting for its completion
- retrieving the location of the vhd file
- copying the vhd to a known storage account blob container as a dependency for compute/disks and compute/images modules

The custom image creation is optional in the dependency pipeline and disabled by default
image

## Type of Change

## Checklist

- [ ] I did format my code


.github/workflows/platform.dependencies.yml
```diff
@@ -286,6 +291,146 @@ jobs:
...
+            $deploymentOutput = '${{ steps.deploy_imgt.outputs.deploymentOutput }}'
+            $imageTemplateName = (ConvertFrom-Json $deploymentOutput).name
+            $imageTemplateResourceGroup = (ConvertFrom-Json $deploymentOutput).resourceGroupName
+            Write-Verbose "imageTemplateName: $imageTemplateName" -Verbose
+            Write-Verbose "imageTemplateResourceGroup: $imageTemplateResourceGroup" -Verbose
+            Write-Output ('::set-output name={0}::{1}' -f 'imageTemplateName', $imageTemplateName)
+            Write-Output ('::set-output name={0}::{1}' -f 'imageTemplateResourceGroup', $imageTemplateResourceGroup)
+          azPSVersion: 'latest'
+
+  job_deploy_vhd_to_sa:
```
MrMCake, line 340: Would it make sense to merge these steps to the image template deployment? Would avoid the need to pass outputs around

utilities/pipelines/dependencies/Microsoft.VirtualMachineImages/imageTemplates/parameters/parameters.json
```diff
@@ -0,0 +1,44 @@
...
+                "sku": "19h2-evd",
+                "version": "latest"
+            }
+        },
+        "customizationSteps": {
+            "value": [
+                {
+                    "type": "WindowsRestart",
+                    "restartTimeout": "30m"
+                }
```
MrMCake, line 37: Do we need a step?

.github/workflows/platform.dependencies.yml
eriqua: Outputs would need to be passed even between steps. The main reasons why I've put it in a separate job are: 

1. It is consistent with KV and storage post deployment which are also in a separate job
2. Having them in the same job would mean running a double step for the agent setup. One as a step in the job and one as a step in the deployment action and I found that a bit messy to follow. Anyway we could discuss about it.

utilities/pipelines/dependencies/Microsoft.VirtualMachineImages/imageTemplates/parameters/parameters.json
eriqua: `customizationSteps` is a required parameter for the ImageTemplate module. I guess the restart is the one having less impact (it's also the one used by the default module parameters)

MrMCake (approved):

PR #1112, opened 2022-03-09
Virtual Network - Subnet Child Resource Remediation

### LINKED ISSUE #1111: Virtual Network module fails when deploying to a network that has existing workloads in it
If the virtual network already contains workloads in it, attempting to redeploy the vnet results in failure.

## Description

The reason is that we are defining the subnet resources outside of the virtual network resource, and that tries to delete all the subnets in the virtual network, which fails. Although we are defining the subnets as a child resource. This doesn't work with this type of resource as all the subnets need to be in the VNET resource.

This was discovered when running the dependency pipeline (in a fork) and had the following error:

```powershell
     | 3:53:28 PM - The deployment 'virtualNetworks-20220309T1503008054Z' failed with error(s). Showing 1 out of 1 error(s).
Status Message: Subnet crawl-az-subnet-x-001 is in use by /subscriptions/20d6fbfe-b049-471c-95af-1369d14d0d45/resourceGroups/validation-rg/providers/Microsoft.Network/networkInterfaces/adp-crawl-vm-01-nic-01/ipConfigurations/ipconfig01 and cannot be deleted. In order to delete the subnet, delete all the resources within the subnet. See aka.ms/deletesubnet. (Code: InUseSubnetCannotBeDeleted)
CorrelationId: b833c292-0cba-4699-91b1-3213358bbb78
```

### Additional references.

Looking up this issue, I can see it already raised here https://github.com/Azure/bicep/issues/4653 with the recommendation to have it in the single vnet resource https://github.com/Azure/bicep/issues/4653#issuecomment-930420476 and this is also still an open item here https://github.com/Azure/azure-quickstart-templates/issues/2786 

### Steps to reproduce

1. Create a VNET using the module
2. Create a workload in that VNET
3. Redeploy the VNET module

### Screenshots

During a deployment, this is what happens inside the VNET (when it doesn't have resources)
image

However, this does solve it (AS A WORKAROUND) and a roll back to the PR https://github.com/Azure/ResourceModules/pull/1081 
image

We need to discuss on:
- Removing the child resource subnet
- Align the subnet properties to the ARM documentation https://docs.microsoft.com/en-us/azure/templates/microsoft.network/virtualnetworks/subnets?tabs=bicep#subnetpropertiesformat
- What are we forgoing from user experience and flexibility if we do that

ahmadabdalla:
# Change

This PR is intended to close #1111 where the virtual network module fails when the vnet contains existing azure workloads. 

Solution:
- Port all subnet properties into the virtual network - subnets array
- remove the custom features (NetworkSecurityGroupResourceGroupName)
- Rename 'Name' to 'Id' for (NetworkSecurityGroup) (RouteTables) (NatGateways)
- Remove the Service Endpoints formatting and variable and align to template reference documentation.
- Added note for the current design for VNET

Notes:
- Porting the features from the subnet child module into the array in the vnet resource

# Testing

[Network: VirtualNetworks](https://github.com/Azure/ResourceModules/actions/workflows/ms.network.virtualnetworks.yml)

Dependency Pipeline (Fork)
image

## Type of Change

## Checklist

- [ ] I did format my code


MrMCake:
Quote from issue: Hey @ahmadabdalla, we should definitely not roll back 1:1 as I removed it one purpose due to an issue that was caused by keeping it (most notably e.g. that the original implementation would clash with ALZ policies). I'd suggest we either find a way to reference 'existing' subnets and pass them in (e.g. with an existing reference if that doesn't throw an exception if empty), or we must duplicate the child resource's property (which would be the worst case scenario).

ahmadabdalla:
Totally agree with not rolling back. Hence it's a draft to get the module working so we can iterate. Pending discussion

MrMCake:
After reading up on it and experiencing the same behavior with Security Rules for NSG we should probably afterall duplicate all properties. I makes sense what the PG says to 'why' it is implemented in the way it is (i.e. enable removal of entries) - it's just very bad news for us & unfortunately not exactly a consistent practice across Azure resources.

ahmadabdalla:
Route tables from my experience has a similar problem for the routes property. If they're set outside of the route table object, they'll get cleared up first. This caused an issue for a customer who had an AKS resource that manages its own routes, and when the pipeline ran, ARM deleted the AKS routes and only created the ones in code. AKS does recover its routes but that's a service disruption. So it's not as  "incremental" as we want on these resources

MrMCake (approved):

PR #1182, opened 2022-03-25
Consolidated ADO and GHA to use a single variables file for environment variables

### LINKED ISSUE #953: Simplify tokenreplacement and variables in the repo
## Description

Written from the PoV of a contributor.

The following scenarios comes to mind:
1. A contributor forks repo, sets up variables and token settings to deploy dependencies, validation and publishing resources. After making a new module or adjusting one, the contributors repo will either need to cherrypick or revert changes in their repo before doing a PR to the Azure/ResourceModules repo. These settings should maybe be moved to something that is not controlled in code?
2. When using the solution to set up a fork for PoC and contribution, i update `.\settings.json`, `.github\variables\global.variables.json` and `.azuredevops\pipelineVariables\global.variables.yml` so it reflects my own setup or align with organization naming conventions. When I then run pipelines most pipelines fail as they are referring to a "hardcoded" rg for validation, `validation-rg`. This is not a true variable for the time being as it requires some search and replace if the RG variables are changed in the variable files.

There are many moving parts to configure a solution for variable management in the solution:
- 3 files that need to change interdependently:
  - `.\settings.json`
  - `.github\variables\global.variables.json`
  - `.azuredevops\pipelineVariables\global.variables.yml`
- A set of secrets for GH
- A variable group + service connection for ADO

Could this be simpler where both the token replacer and variables are managed from a single place, which does not affect the content of the git repo, making also the contribution aspect easier?

ahmadabdalla:
Closes #953 

Moved to a common and consolidated `global.variables.yml` file that works for both GHA and ADO.

## Key Tasks
- Removed duplicates and have a single location to update environment variables for both CI platforms.
- Maintain existing native capability of ADO to use YAML files for variable templates.
- Created a new composite GHA that reads a YAML list containing key value pairs, like the ADO variable file. With two tasks:
  - Uses the existing 'Setup Agent' feature to install the `powershell-yaml` module from the PowerShell gallery
  - Reads the variable YAML file using a newly developed generic script `Add-YamlListToFile` that deserializes the YAML List into a key value pair that is outputted to a file
 
### GitHub Module Workflows
- new Env: Variable that point to the path of the 'global.variables.yml`.
- Replace the marketplace action `deep-mm/set-variables@v1.0` with our new composite action.
- Move away from the variables that use the word `default` so that we align both CI platforms.

### Azure DevOps Module Pipelines
- Change the path of the template variable to the new root path so that the file works for both ADO and GHA.
- Move away from the variables that use the word `default` so that we align both CI platforms.

### Global Variable File
- Reorganized the variables and added sections to help distinguish the two CI platforms specific variables
- Removed the word `default` to that we align both CI platforms.
- Removed the ADO `modulePath: arm/$(moduleName)` variable as it causes conflict with GHA, but not impacting ADO.

## Design

image

## Testing
- Fully tested for GHA and ADO Module Pipeline and Dependency Pipeline

[Authorization: PolicyAssignments](https://github.com/Azure/ResourceModules/actions/workflows/ms.authorization.policyassignments.yml)

image


.github/actions/templates/setEnvironmentVariables/action.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: |
+        # Grouping task logs
+        Write-Output "::group::Setup agent"
+
+        # Load used functions
+        . (Join-Path $env:GITHUB_WORKSPACE 'utilities' 'pipelines' 'sharedScripts' 'Set-EnvironmentOnAgent.ps1')
+
+        # Define PS modules to install on the runner
+        $Modules = @(
+            @{ Name = 'powershell-yaml'; Version = '0.4.2'}
```
eriqua, line 48: @MrMCake @MariusStorhaug @ahmadabdalla at the beginning of the initiative we discussed about leveraging this external module a couple of times, eventually deciding not to use that, since we wanted to avoid external dependencies. This is the reason why the getInputParameters action is the way it is. Do we want to rediscuss this?

global.variables.yml
```diff
@@ -1,4 +1,37 @@
 variables:
+  ###########################################################################################################################
```
eriqua, line 2: Can we still use ADO variable groups with this change?

.azuredevops/platformPipelines/platform.dependencies.yml
```diff
@@ -24,7 +24,7 @@ trigger: none
 #       - 'utilities/pipelines/dependencies/**'
 
 variables:
-  - template: '/.azuredevops/pipelineVariables/global.variables.yml'
+  - template: '../../global.variables.yml'
```
MrMCake, line 27: Shouldn't this work too for all references?

```suggestion
  - template: './global.variables.yml'
```

global.variables.yml
ahmadabdalla: Yes. It's not impacting existing ADO capabilities

.github/actions/templates/setEnvironmentVariables/action.yml
MrMCake: It was, however with the variable file we now have a more complicated construct to deal with (compared to the runtime parameters of the pipeline). Maybe we need to revisit this decision. Let's have a discussion some time.

.github/actions/templates/setEnvironmentVariables/action.yml
```diff
@@ -0,0 +1,71 @@
...
+      id: set-env-variables
+      shell: pwsh
+      run: |
+        # Grouping task logs
+        Write-Output "::group::Set environment variables"
+
+        # Load used functions
+        . (Join-Path $env:GITHUB_WORKSPACE 'utilities' 'pipelines' 'sharedScripts' 'Add-YamlListToFile.ps1')
+
+        # Convert YAML Variable File to Environment Variables
```
MrMCake, line 68: I'd appreciate if we could add the 'Invoked task with' snippet too

global.variables.yml
eriqua: Shouldn't ADO variables be declared with 
```
- name: ..
  value: ..
``` 
to support variable groups? Just trying to fully understand the implications of this. If that is still the case, we should probably mention that in the docs. For now we only have one variable group declared in all ADO pipelines, i.e. 
```
- group: 'PLATFORM_VARIABLES'
``` 
the one containing pipeline secrets. Similarly, in case other groups are needed going forward, they cannot be added to the global.variables.yml file. They need to be added to pipelines instead. Would you agree?

ahmadabdalla: Not sure I understand the question. From ADO perspective, we continue to use the same methods, whether its pipeline variables, variable groups or variables in templates. Only change is the path of the template YAML to live outside of the ADO folder

.azuredevops/platformPipelines/platform.dependencies.yml
ahmadabdalla: I had that before and the pipelines fail to start because the starting path is where the pipeline is stored. Error:


/.azuredevops/modulePipelines/ms.authorization.policyassignments.yml: Could not find /.azuredevops/modulePipelines/global.variables.yml in repository self hosted on https://github.com/ using commit a84a6d28484dc3c02f100c737950aedbf6a2340d. GitHub reported the error, "Not Found"

.github/actions/templates/setEnvironmentVariables/action.yml
ahmadabdalla: done

.azuredevops/platformPipelines/platform.dependencies.yml
MrMCake: Alright then. A shame GitHub & Azure DevOps don't work the same

docs/wiki/Getting started - Setup environment.md
```diff
@@ -163,14 +163,14 @@ To use the environment's pipelines you should use the information you gathered d
 
 ### 3.2.2 Setup variables file
 
-The primary pipeline variable file `.github/variables/global.variables.json` hosts the fundamental pipeline configuration. In the file you will find and can configure settings such as:
+The primary pipeline variable file `global.variables.yml` hosts the fundamental pipeline configuration. In the file you will find and can configure settings such as:
```
MrMCake, line 166: @eriqua, we should be aware this may require updates in the updated Wiki too

MrMCake: Same for the other files like `PipelineDesign`

MrMCake: Actually nevermind. This page here is already new. Then only in the `PipelineDesign`.

global.variables.yml
MrMCake: As discussed: As long as we don't mix the short syntax 

```yaml
<key>: <value>
```

with the 'long syntax'

```yaml
- name: <key>
  value: <value>
```
we should be fine

utilities/pipelines/sharedScripts/Add-YamlListToFile.ps1
```diff
@@ -0,0 +1,87 @@
...
+        # Validate Input and Output Path
+        try {
+            if (-not ($InputFileContent = Get-Content -Path $InputFilePath)) {
+                throw "Invalid Input File Path: $InputFilePath"
+            }
+            if (-not (Test-Path $OutputFilePath)) {
+                throw "Invalid Output File Path: $OutputFilePath"
+            }
+        } catch {
+            throw $PSitem.Exception.Message
```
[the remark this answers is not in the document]
MrMCake: Same for below

global.variables.yml
eriqua: Yes, addressed offline today, we can solve it.

utilities/pipelines/sharedScripts/Add-YamlListToFile.ps1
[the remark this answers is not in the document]
ahmadabdalla: thanks for the note. Addressed

docs/wiki/Getting started - Setup environment.md
ahmadabdalla: @MrMCake Is there something need to action here?

MrMCake (approved):

PR #1192, opened 2022-03-29
Updated template scope determination to be more robust

MrMCake:
# Change

- Updated template scope determination to be more robust

## Type of Change

### DIFF
--- a/utilities/pipelines/sharedScripts/Get-ScopeOfTemplateFile.ps1
+++ b/utilities/pipelines/sharedScripts/Get-ScopeOfTemplateFile.ps1
@@ -30,7 +30,7 @@ function Get-ScopeOfTemplateFile {
     if ((Split-Path $templateFilePath -Extension) -eq '.bicep') {
         # Bicep
         $bicepContent = Get-Content $templateFilePath
-        $bicepScope = $bicepContent | Where-Object { $_ -like '*targetscope =*' }
+        $bicepScope = $bicepContent | Where-Object { $_ -match '^targetscope = .+' }
         if (-not $bicepScope) {
             $deploymentScope = 'resourcegroup'
         } else {


eriqua (commented):
Looks good. Since this will be an easy way to trigger all pipelines I'd suggest to merge it right after the readme updates if you agree

utilities/pipelines/sharedScripts/Get-ScopeOfTemplateFile.ps1
```diff
@@ -30,7 +30,7 @@ function Get-ScopeOfTemplateFile {
     if ((Split-Path $templateFilePath -Extension) -eq '.bicep') {
         # Bicep
         $bicepContent = Get-Content $templateFilePath
-        $bicepScope = $bicepContent | Where-Object { $_ -like '*targetscope =*' }
+        $bicepScope = $bicepContent | Where-Object { $_ -match '^targetscope = .+' }
```
khaliddermoumi, line 33: Hi! This code will still fail with lines such as "targetScope = 'managementGroup' // some comment",
also with "targetScope =    'managementGroup'" (-> more blanks than expected, etc.):

```powershell
        $bicepContent = Get-Content $templateFilePath
        $bicepScope = $bicepContent | Where-Object { $_ -match '^targetscope = .+' }
        if (-not $bicepScope) {
            $deploymentScope = 'resourcegroup'
        } else {
            $deploymentScope = $bicepScope.ToLower().Split('=')[-1].Replace("'", '').Trim()
        }
```
I believe it is better to use regex captures here, and be a little more strict about the regex pattern. This would result in something like:
```powershell
        $bicepContent = Get-Content $templateFilePath -Raw
        $bicepScopeMatch = [regex]::Match($bicepContent, '(?m)^\s*targetScope\s*=\s*''(\S+)''')
        if (-not $bicepScopeMatch.Success) {
            $deploymentScope = 'resourcegroup'
        } else {
            $deploymentScope = $bicepScopeMatch.Captures.Groups[1].Value
        }
```
(the Flag " -Raw" is required in the first line, to read the file content to a string instead of a string array).

khaliddermoumi (changes requested):

utilities/pipelines/sharedScripts/Get-ScopeOfTemplateFile.ps1
khaliddermoumi: BTW: this is just a proposal, feel free to adopt it or not.
I believe the script just work with everything which is legal bicep. A unit test would also make sense here I believe.

MrMCake: Nice. Would you mind creating a PR @khaliddermoumi ? Then I'd close this one. It's your solution after all ;)

khaliddermoumi: @MrMCake happy to do so! :)

PR #1257, opened 2022-04-12
Fixed GitHub parameter fetch

MrMCake:
# Change

- The compression used for passing the parameter files from one job to another auto-converted arrays into single strings that broke the consumption via the matrix. This is now fixed.

Pipeline reference
[DesktopVirtualization: HostPools](https://github.com/Azure/ResourceModules/actions/workflows/ms.desktopvirtualization.hostpools.yml)

Re-test with multi-param example
[AnalysisServices: Servers](https://github.com/Azure/ResourceModules/actions/workflows/ms.analysisservices.servers.yml)

## Type of Change

### DIFF
--- a/.github/actions/templates/getParameterFiles/action.yml
+++ b/.github/actions/templates/getParameterFiles/action.yml
@@ -28,7 +28,7 @@ runs:
         Write-Verbose "Invoke task with" -Verbose
         Write-Verbose ($functionInput | ConvertTo-Json | Out-String) -Verbose
         # Get the list of parameter file paths
-        $parameterFilePaths = Get-ModuleParameterFiles @functionInput -Verbose
+        [array]$parameterFilePaths = Get-ModuleParameterFiles @functionInput -Verbose
         # Output values to be accessed by next jobs
         Write-Output ('::set-output name=parameterFilePaths::{0}' -f ($parameterFilePaths | ConvertTo-Json -Compress))
         Write-Output "::endgroup::"
--- a/.github/actions/templates/getParameterFiles/action.yml
+++ b/.github/actions/templates/getParameterFiles/action.yml
@@ -30,5 +30,7 @@ runs:
         # Get the list of parameter file paths
         [array]$parameterFilePaths = Get-ModuleParameterFiles @functionInput -Verbose
         # Output values to be accessed by next jobs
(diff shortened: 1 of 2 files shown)


arm/Microsoft.Web/sites/.parameters/fa.parameters.json
```diff
@@ -23,7 +23,7 @@
             "value": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Storage/storageAccounts/adp<<namePrefix>>azsafa001"
         },
         "functionsWorkerRuntime": {
-            "value": "powershell"
+            "value": "<<tenantId>>"
```
eriqua, line 26: is this intended?

.github/actions/templates/getParameterFiles/action.yml
```diff
@@ -30,5 +30,10 @@ runs:
         # Get the list of parameter file paths
         $parameterFilePaths = Get-ModuleParameterFiles @functionInput -Verbose
         # Output values to be accessed by next jobs
-        Write-Output ('::set-output name=parameterFilePaths::{0}' -f ($parameterFilePaths | ConvertTo-Json -Compress))
+        $compressedOutput = $parameterFilePaths | ConvertTo-Json -Compress
+        if($compressedOutput -notmatch "\[.*\]") {
+          $compressedOutput = "[$compressedOutput]"
+        }
+        Write-Verbose "Publishing output: $compressedOutput" -Verbose
+        Write-Output "::set-output name=parameterFilePaths::$compressedOutput"
```
eriqua, line 38: As you teach me 
```suggestion
        Write-Output ('::set-output name=parameterFilePaths::{0}' -f $compressedOutput)
```

arm/Microsoft.Web/sites/.parameters/fa.parameters.json
MrMCake: Nope

.github/actions/templates/getParameterFiles/action.yml
MrMCake: If we have more than one parmaeter - or - we'd need to access a property of the parameter I'd definitely agree. In this case however I'd suggest to use the 'easy' syntax.

eriqua (approved):
🆗

PR #1298, opened 2022-04-14
Users/jozessi/updateaml

dr-dolittle:
# Description

Update identity enablement from bool to actual object supporting user assigned managed identities

Add computes child resource

Update readmes

Add sections to bicep files

## Pipeline references

[MachineLearningServices: Workspaces](https://github.com/quervernetzt/ResourceModules/actions/workflows/ms.machinelearningservices.workspaces.yml)

# Type of Change

# Checklist

- [x] I did format my code


arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
```diff
@@ -31,8 +34,8 @@ param associatedContainerRegistryResourceId string = ''
 @description('Optional. Specify the type of lock.')
 param lock string = 'NotSpecified'
 
-@description('Optional. Enables system assigned managed identity on the resource.')
-param systemAssignedIdentity bool = false
+@description('Optional. Identity for the resource.')
+param identity object = {}
```
MrMCake, line 38: Is there a reason why you changed this? We use the same implementation (almost) across the board and if the current one does not serve all use cases, we should update the other ones too. cc: @MariusStorhaug

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
```diff
@@ -0,0 +1,118 @@
...
+param computeType string
+
+@description('Optional. The properties of the compute. Will be ignored in case "computeResourceId" is set.')
+param computeProperties object = {}
+
+@description('Optional. Enable telemetry via the Customer Usage Attribution ID (GUID).')
+param enableDefaultTelemetry bool = true
+
+// =========== //
+// Existing resources //
```
MrMCake, line 65:
```suggestion
// Existing resource //
```
```suggestion
// Existing resources references // 
```

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
```diff
@@ -0,0 +1,118 @@
...
+  }
+  parent: machineLearningWorkspace
+  identity: identity
+  properties: union({
+    computeLocation: computeLocation
+    description: computeDescription
+    disableLocalAuth: computeDisableLocalAuth
+    computeType: computeType
+  }, (!empty(computeResourceId) ? {
+    resourceId: computeResourceId
```
MrMCake, line 102: If you'd agree we should try to stay aligned with the Azure resource reference. The rationale is, that it is (in theory) easier for consumers, as our readMe would be closer aligned to the offical Azure Resource Reference. If we diverge from it, there should be a good reason. 
> Note: This comment also relates to other properties like `computeDescription`

```suggestion
    resourceId: resourceId
```

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
```diff
@@ -153,6 +160,25 @@ resource workspace 'Microsoft.MachineLearningServices/workspaces@2021-04-01' = {
...
+  params: {
+    machineLearningWorkspaceName: workspace.name
+    name: compute.name
+    location: location
+    sku: sku
+    identity: identity
+    tags: tags
+    deployCompute: compute.deploy
+    computeLocation: compute.location
+    computeDescription: compute.description
```
MrMCake, line 174: If you specify it like this, `description` will become a required parameter even though the child module does not declare it as such.

The recommendation is (and the same accounts for all other properties): If a parameter is optional, use the syntax: `description: contains(compute, 'description') ? compute.description :  ''`
In words: If the compute object contains the property, use it, otherwise use the default specified in the child itself.

Unfortunately, there is no feature in bicep that it would automatically pick up the child default (like a special keyword)

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
dr-dolittle: @MrMCake  @MariusStorhaug The way it was implemented it supported only system assigned managed identity. This way it is possible to provide user assigned managed identities as well. I propose to keep it this way (or similar) and if necessary also change it in the other modules. What do you think?

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
dr-dolittle: Updated

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
dr-dolittle: @MrMCake Updated. "description" is not possible -> "This expression is referencing its own declaration, which is not allowed.", therefore I use "computeDescription".

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
dr-dolittle: @MrMCake Updated.

MariusStorhaug:
Could you check any other modules that supports user assigned identity, that would be where you see one way of supporting both. One thing we could work to improve, is the user assigned identity object that is created. Should really only be an array of strings.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
MariusStorhaug: Could you check any other modules that supports user assigned identity, that would be where you see one way of supporting both. One thing we could work to improve, is the user assigned identity object that is created. Should really only be an array of strings.

dr-dolittle: @MariusStorhaug I changed the implementation following your suggestion. Is it ok that way?

MariusStorhaug: Ye, the implementation seems similar to the other ones now. I see in the diff that when we did change this last time, the resource type did not support `userAssignedIdentities`, now it seems it does. Thank you for adding it and help keeping the code consistent across modules!

arm/Microsoft.MachineLearningServices/workspaces/.parameters/parameters.json
```diff
@@ -17,6 +17,34 @@
         "associatedApplicationInsightsResourceId": {
             "value": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Insights/components/adp-<<namePrefix>>-az-appi-x-001"
         },
+        "systemAssignedIdentity": {
+            "value": true
+        },
+        "computes": {
```
MariusStorhaug, line 23: Lets add it to the deployment validation too:

```suggestion
        },
        "userAssignedIdentities": {
            "value": {
                "/subscriptions/<<subscriptionId>>/resourcegroups/validation-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/adp-<<namePrefix>>-az-msi-x-001": {}
            }
        },
        "computes": {
```

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
```diff
@@ -153,6 +171,26 @@ resource workspace 'Microsoft.MachineLearningServices/workspaces@2021-04-01' = {
   }
 }
 
+module workspace_computes 'computes/deploy.bicep' = [for compute in computes: {
+  name: '${workspace.name}-${compute.name}-compute'
+  params: {
+    machineLearningWorkspaceName: workspace.name
+    name: compute.name
+    location: location
+    sku: sku
```
MariusStorhaug, line 180: Do we want this to rely on the values from within the compute object that you iterate through or should it be inherited from the parent?

```suggestion
    sku: compute.sku
```

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
```diff
@@ -153,6 +171,26 @@ resource workspace 'Microsoft.MachineLearningServices/workspaces@2021-04-01' = {
...
+    systemAssignedIdentity: systemAssignedIdentity
+    userAssignedIdentities: userAssignedIdentities
```
MariusStorhaug, line 182: Do we want this to rely on the values from within the compute object that you iterate through or should it be inherited from the parent?

```suggestion
    systemAssignedIdentity: compute.systemAssignedIdentity
    userAssignedIdentities: compute.userAssignedIdentities
```

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
MrMCake: @dr-dolittle actually it is. Use the annotation `sys.description()` instead. We have the same in case of e.g. the resources in the `Microsoft.Authorization` provider namespace

arm/Microsoft.MachineLearningServices/workspaces/.parameters/parameters.json
```diff
@@ -17,6 +17,34 @@
...
+            "value": [
+                {
+                    "name": "DefaultCPU",
+                    "location": "westeurope",
+                    "description": "Default CPU Cluster",
+                    "disableLocalAuth": false,
+                    "deploy": true,
+                    "type": "AmlCompute",
```
MariusStorhaug, line 31: Lets discuss this point under the suggestion on the deploy.bicep file for the workspaces module where you call the compute module. Come back here if you want to treat these independent from the parent settings.

```suggestion
                    "type": "AmlCompute",
                    "systemAssignedIdentity": true,
                    "userAssignedIdentities": {
                        "/subscriptions/<<subscriptionId>>/resourcegroups/validation-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/adp-<<namePrefix>>-az-msi-x-001": {}
                    },
```

MariusStorhaug (changes requested):

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
```diff
@@ -0,0 +1,132 @@
...
+// Outputs     //
+// =========== //
+@description('The name of the compute.')
+output name string = machineLearningWorkspaceCompute.name
+
+@description('The resource ID of the compute.')
+output resourceId string = machineLearningWorkspaceCompute.id
+
+@description('The resource group the compute was deployed into.')
+output resourceGroupName string = resourceGroup().name
```
eriqua, line 132: I'd suggest to add this output and regenerate the readme file
```suggestion
output resourceGroupName string = resourceGroup().name

@description('The principal ID of the system assigned identity.')
output systemAssignedPrincipalId string = systemAssignedIdentity && contains(machineLearningWorkspaceCompute.identity, 'principalId') ? machineLearningWorkspaceCompute.identity.principalId : ''
```

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
dr-dolittle: @MrMCake Is updated.

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
dr-dolittle: @eriqua Is updated.

arm/Microsoft.MachineLearningServices/workspaces/.parameters/parameters.json
dr-dolittle: @MariusStorhaug Is updated.

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
dr-dolittle: @MariusStorhaug Updated.

arm/Microsoft.MachineLearningServices/workspaces/.parameters/parameters.json
dr-dolittle: @MariusStorhaug Updated.

arm/Microsoft.MachineLearningServices/workspaces/deploy.bicep
dr-dolittle: @MariusStorhaug Updated.

dr-dolittle:
@MrMCake @MariusStorhaug @eriqua Thanks for your feedback. I implemented your changes as suggested. There are two more open topics: in case of "create" it is actually possible to deploy multiple times without error, in case of attach it fails. Also, I want to clarify which properties are actually required for attach versus create. I will update it as soon as I know more.

dr-dolittle:
> @MrMCake @MariusStorhaug @eriqua Thanks for your feedback. I implemented your changes as suggested. There are two more open topics: in case of "create" it is actually possible to deploy multiple times without error, in case of "attach" it fails. Also, I want to clarify which properties are actually required for "attach" versus "create". I will update it as soon as I know more.

@MrMCake @MariusStorhaug @eriqua Is now adjusted and fine from my perspective. Please review.

But the pipeline fails dues to failing static validation -> https://github.com/quervernetzt/ResourceModules/actions/runs/2179547321:

image
image

dr-dolittle:
> > @MrMCake @MariusStorhaug @eriqua Thanks for your feedback. I implemented your changes as suggested. There are two more open topics: in case of "create" it is actually possible to deploy multiple times without error, in case of "attach" it fails. Also, I want to clarify which properties are actually required for "attach" versus "create". I will update it as soon as I know more.
> 
> @MrMCake @MariusStorhaug @eriqua Is now adjusted and fine from my perspective. Please review.
> 
> But the pipeline fails dues to failing static validation -> https://github.com/quervernetzt/ResourceModules/actions/runs/2179547321:
> 
> image
> 
> image
> 
> Can you please check?
> 
> Thanks

@MrMCake This one also fails -> https://github.com/quervernetzt/ResourceModules/actions/runs/2189802004

image

dr-dolittle:
@MrMCake Pipeline run was successful. Can we merge now? Thanks

arm/Microsoft.MachineLearningServices/workspaces/.parameters/parameters.json
```diff
@@ -17,6 +17,53 @@
...
+                            "maxNodeCount": 3,
+                            "minNodeCount": 0,
+                            "nodeIdleTimeBeforeScaleDown": "PT5M"
+                        },
+                        "vmPriority": "Dedicated",
+                        "vmSize": "STANDARD_DS11_V2"
+                    }
+                }
+                // {
+                //     "name": "DefaultAKS",
```
MrMCake, line 57: I guess this is commented because it cannot be deployed in parallel?

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
```diff
@@ -0,0 +1,136 @@
...
+    mode: 'Incremental'
+    template: {
+      '$schema': 'https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#'
+      contentVersion: '1.0.0.0'
+      resources: []
+    }
+  }
+}
+
+resource machineLearningWorkspaceCompute 'Microsoft.MachineLearningServices/workspaces/computes@2022-01-01-preview' = if (deployCompute == true) {
```
MrMCake, line 101: Why the condition?

arm/Microsoft.MachineLearningServices/workspaces/.parameters/parameters.json
dr-dolittle: @MrMCake The reason why it is commented is to show how it would work but it is not enabled as it would require an AKS cluster. Do we have an AKS as dependency available?

It is possible to deploy multiple compute resources in parallel.

arm/Microsoft.MachineLearningServices/workspaces/computes/deploy.bicep
dr-dolittle: @MrMCake As written in the README doing a deployment for an attached compute (e.g. AKS) multiple times will fail, i.e. the first one will succeed and attach the resource, but as the operation is not idempotent subsequent resources will fail. So, when you attach a compute you need to set the flag to false after the first deployment. For created resources via AML this flag is not relevant as the operation is idempotent. The PG is aware of the problem but so far has not resolved it.

MrMCake: ah right, my bad

arm/Microsoft.MachineLearningServices/workspaces/.parameters/parameters.json
MrMCake: Not yet, no. But we can do this in the future with a new dependency approach. I'm usually not a big fan of these kind of comments - so could you please add a smal disclaimer on top if it saying : `This is an example of a link to an existing AKS resource`

dr-dolittle: @MrMCake I removed the commented part.

MrMCake (approved):

MariusStorhaug (approved):

PR #1318, opened 2022-04-19
Updating pipeline triggers

eriqua:
# Description

Including arm/.global folder changes to pipeline triggers for both GH and ADO

| Pipeline |
| - |
| |

# Type of Change

# Checklist

- [ ] I did format my code

### DIFF
--- a/.azuredevops/modulePipelines/ms.authorization.policysetdefinitions.yml
+++ b/.azuredevops/modulePipelines/ms.authorization.policysetdefinitions.yml
@@ -22,7 +22,7 @@ trigger:
       - '/.azuredevops/modulePipelines/ms.authorization.policysetdefinitions.yml'
       - '/.azuredevops/pipelineTemplates/module.*.yml'
       - '/arm/Microsoft.Authorization/policySetDefinitions/*'
-      - '/arm/.global/**'
+      - '/arm/.global/global.module.tests.ps1'
     exclude:
       - '/**/*.md'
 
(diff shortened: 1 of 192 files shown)


.azuredevops/modulePipelines/ms.authorization.policysetdefinitions.yml
```diff
@@ -22,6 +22,7 @@ trigger:
       - '/.azuredevops/modulePipelines/ms.authorization.policysetdefinitions.yml'
       - '/.azuredevops/pipelineTemplates/module.*.yml'
       - '/arm/Microsoft.Authorization/policySetDefinitions/*'
+      - '/arm/.global/**'
```
MrMCake, line 25: everything in global? I'm just wondering as we have also dependency templates in that folder (in the dependency PoC). And not every module will use them - but would be triggered regardless

eriqua: True, although I now wonder if anything not having a global impact should be included in a folder named global

MrMCake: A good point. One could however, also argue that `global` = `shared` and not every module has to consume everything that is shared. This is some deep philosophical conversation right here

eriqua: I'm not even sure why the .global folder is in the arm folder actually. I think this should be discussed as well. Isn't it a utility pipeline afterall?

MrMCake: It is there as it is module specific. In any case I'd suggest to open this as a new discussion and keep separate from this PR. For the time being I'd anyhow strongly suggest to be more specific than `global*`

eriqua: This would be listing all files contained in the .global folder right now. Fine but hopefully temporary 😆

MrMCake: Aren't we after the `global.tests.ps1`? :D

eriqua: and related helper. current and future.

Issue #1154, opened 2022-03-16
Add PIP to Load Balancer


rahalan:
## Description

Load Balancer is not creating a PIP on their own (if not provided), as other modules do.

jontreynes:
The FW expects a already deployed Public IP for the IpConfiguration. Other modules, like the VM, are able to create their own, if none is provided. The FW should create one, if none is provided

jontreynes:
it looks like this is the logic that is being used in other modules
https://github.com/Azure/ResourceModules/blob/main/arm/Microsoft.Network/natGateways/deploy.bicep#L136

jontreynes:
Trying to see if this another place where we default creating the PIP

The logic here translates to
https://github.com/Azure/ResourceModules/blob/80087b330c2eae20e720a2d53cf24af2c90478a8/arm/Microsoft.Compute/virtualMachines/.bicep/nested_networkInterface.bicep#L41

if (pipconfiguration has publicIPPrefixId){
	if (!publicIPPrefixId)
		publicIPPrefixId = ''
	else
		publicIPPrefixId = ipConfiguration.pipconfiguration.publicIPPrefixId
}
else
	publicIPPrefixId = ''

jontreynes:
should reference pip module

jontreynes:
bastionhost has this logic

module publicIPAddress '.bicep/nested_publicIPAddress.bicep' = if (empty(publicIPAddressId)) {

jontreynes:
[![Network: AzureFirewalls](https://github.com/Azure/ResourceModules/actions/workflows/ms.network.azurefirewalls.yml/badge.svg)](https://github.com/Azure/ResourceModules/actions/workflows/ms.network.azurefirewalls.yml)

jontreynes:
So adding an existing public ip address already exists under the ipconfigurations parameter where you can provide a list of value objects like

```
            "value": [
                {
                    "name": "ipConfig01",
                    "publicIPAddressResourceId": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/publicIPAddresses/adp-<<namePrefix>>-az-pip-x-fw",
                    "subnetResourceId": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-azfw/subnets/AzureFirewallSubnet"
                }
```

[...]

Issue #401, opened 2021-11-04
Improve test files to cover more test cases


rahalan:
- create for each module min, max and scenario parameter files to cover more testcases and test the deployment of every module aspect

- add checklist to PR template, to maintain parameter files (if new parameter is added)

rahalan:
Split the issue into issues per namespace

jontreynes:
I'm updating just the Azurefirewall one right now while I was updating it here #1328

AlexanderSehr:
As modules are moved and partly enforced in AVM, closing this

PR #1328, opened 2022-04-21
1154 Add Public IP to Azure Firewall

jontreynes:
# Description

The work completes a portion of #1154 in which I added logic to create a public ip if one is not provided for Azure Firewall. I copied the logic from arm\Microsoft.Network\bastionHosts\deploy.bicep. When reusing this, I spotted an unnecessary piece of logic in bastionHosts that creates a resource reference to the existing public ip address if provided. This resource reference was deleted because it was not used. 

I also renamed some components of the firewall to use the abbreviations provided by CAF here https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-abbreviations#networking when defaulting a naming on a child resource in the azurefirewall module

> Note: the dependent resources like vnet id and an existing public ip still contain the nonCAF naming convention like so
```
        "vNetId": {
            "value": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-azfw"
        },
        "publicIPAddressId": {
            "value": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/publicIPAddresses/adp-<<namePrefix>>-az-pip-x-fw"
```

For the sake of completing the work, I left the dependent names alone and just updated the actual firewall name

I also introduced a min.parameters.json file for just azurefirewall - which is a portion of this task #401 

# Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update (Wiki)

> This is a breaking change because I deleted the ipconfigurations parameter and separated it out into vnetId parameter, publicIpAddressId, and/or publicIPAddressObject

# Checklist

Checking this box after checking #401 has not been started
- [x] I did format my code


arm/Microsoft.Network/azureFirewalls/readme.md
```diff
@@ -18,14 +18,15 @@ This module deploys a firewall.
...
+| `Microsoft.Network/publicIPAddresses` | 2021-05-01 |
 
 ## Parameters
 
 **Required parameters**
 | Parameter Name | Type | Description |
 | :-- | :-- | :-- |
-| `ipConfigurations` | array | List of IP Configurations. |
 | `name` | string | Name of the Azure Firewall. |
+| `vNetId` | string | Shared services Virtual Network resource identifier |
```
MrMCake, line 29:
```suggestion
| `vNetId` | string | Shared services Virtual Network resource ID |
```

arm/Microsoft.Network/azureFirewalls/deploy.bicep
```diff
@@ -15,6 +15,15 @@ param azureSkuName string = 'AZFW_VNet'
 ])
 param azureSkuTier string = 'Standard'
 
+@description('Required. Shared services Virtual Network resource identifier')
```
MrMCake, line 18:
```suggestion
@description('Required. Shared services Virtual Network resource ID')
```

arm/Microsoft.Network/azureFirewalls/deploy.bicep
```diff
@@ -102,18 +108,6 @@ param diagnosticMetricsToEnable array = [
   'AllMetrics'
 ]
 
-var ipConfigurations_var = [for ipConfiguration in ipConfigurations: {
```
MrMCake, line 105: Can there only be one IP configuration per firewall?

It used to be an array (and the property still is) but we only have one entry.

arm/Microsoft.Network/azureFirewalls/readme.md
```diff
@@ -47,6 +48,8 @@ This module deploys a firewall.
 | `lock` | string | `'NotSpecified'` | `[CanNotDelete, NotSpecified, ReadOnly]` | Specify the type of lock. |
 | `natRuleCollections` | array | `[]` |  | Collection of NAT rule collections used by Azure Firewall. |
 | `networkRuleCollections` | array | `[]` |  | Collection of network rule collections used by Azure Firewall. |
+| `publicIPAddressId` | string | `''` |  | Specifies the resource ID of the existing public IP to be leveraged by Azure Firewall. If not provided, an Azure Public IP will be created |
+| `publicIPAddressObject` | object | `{object}` |  | Specifies the properties of the public IP to create and be used by Azure Firewall. If it's not provided and publicIPAddressId is empty, a '-pip' suffix will be appended to the Firewall's name. |
```
MrMCake, line 52: If we have this object here we should also provide a `Parameter Usage` example below

arm/Microsoft.Network/azureFirewalls/deploy.bicep
MrMCake: I assume this comes down to the creation of the public IP that would otherwise have been provided in the array?

jontreynes: The template shows there can be many ip configurations per firewall according to the ipconfigurations object which can take an array like this
image

https://docs.microsoft.com/en-us/azure/templates/microsoft.network/2021-05-01/azurefirewalls?tabs=bicep#template-format

so can the bastionhosts which is where I copied the logic from
https://docs.microsoft.com/en-us/azure/templates/microsoft.network/2021-05-01/bastionhosts?tabs=bicep#template-format

When I look at the portal, you can add many public ips
image

However, there's not a portal view for assigning multiple subnets to an azurefirewall
image
I can test this locally if it'll complain.

jontreynes: After I test, we can see if it's worth merging ipconfigurations with the default public ip created in the module

jontreynes: Update: It's not possible to have more than one subnet reference
image

You can have multiple public ips, but I need to check if the CAF documentation is against maintaining multiple public ips for one resource

jontreynes: The general microsoft documentation say's there's use cases for DNAT and SNAT here
https://docs.microsoft.com/en-us/azure/firewall/deploy-multi-public-ip-powershell#:~:text=Azure%20Firewall%20with%20multiple%20public%20IP%20addresses%20is,and%20remove%20public%20IP%20addresses%20for%20Azure%20Firewall.

jontreynes:
moving to draft - going to make some updates like reintroducing ip_configurations_var

arm/Microsoft.Network/azureFirewalls/deploy.bicep
jontreynes: Note: you MUST provide a public ip resource id when providing a subnet Resource Id based on this test

```
![image](https://user-images.githubusercontent.com/19974635/166552796-40f3e649-5394-428b-926d-4af6b8c6fea1.png)

```

So if a subnet id address is provided without a public ip, then the public ip address should be applied to that subnet id. Also, there cannot be multiple subnet ids as shown in the previous example - so I need to create a parser that finds the first subnet id in ipconfigurations and merges the public ip address to create to that object

jontreynes: Also you cannot deploy without a subnet in ipconfigurations
image

jontreynes: and you're required to have at least one pip
image

jontreynes: image


This is what happens when you do 
```
          publicIPAddress: {
            //Use existing public ip, new public ip created as in this module, or none if isCreateDefaultPublicIP is false
            id: !(empty(azureFirewallSubnetPublicIpId)) ? azureFirewallSubnetPublicIpId : (isCreateDefaultPublicIP ? publicIPAddress.outputs.resourceId : null)
          }

```

But it works when you completely take out the publicIpAddress object

So, I need to find out a way to build the ipconfigurations with publicIpAddress and without dependeing on isCreateDefaultPublicIp

Issue #877, opened 2022-01-11
Event Hub namespace default name Bug Report


eriqua:
## Description

The event hub module generates a unique random name in case the name is not passed as a parameter. That is tested with the parameter file min.parameters.json.

The logic creating the unique name must comply with the eventhub namespace naming convention. In particular, the unique name must start with a letter. The current logic allows the generated unique name to start with a number and in that case the deployment fails.
> Example: https://github.com/Azure/ResourceModules/runs/4776425099?check_suite_focus=true#step:4:430

### Steps to reproduce

1. Run the event hub module until the logic generating the unique event hub namespace name creates a string starting with a number
2. The pipeline testing the min.parameters.json fails

### Screenshots

![image](https://user-images.githubusercontent.com/56914614/148963102-fe0f22ee-331e-423a-b002-8e131d3b24d2.png)

rahalan:
Or discuss if we remove the auto-generate name functionality

PR #1334, opened 2022-04-22
Event Hub namespace default name Bug Report #877

arnoldna:
# Description
Updated logic to autogenerate unique default EventHub Namespace name.

## Pipeline references

| Pipeline |
| - |
| [EventHub: Namespaces](https://github.com/arnoldna/ResourceModules/actions/workflows/ms.eventhub.namespaces.yml)|

# Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist

- [ ] I did format my code


arm/Microsoft.EventHub/namespaces/deploy.json
```diff
@@ -0,0 +1,1610 @@
+{
```
MrMCake, line 1: Please remove the compiled bicep template

arm/Microsoft.EventHub/namespaces/deploy.bicep
```diff
@@ -127,7 +127,7 @@ param diagnosticMetricsToEnable array = [
 ]
 
 var maxNameLength = 50
-var uniqueEventHubNamespaceUntrim = uniqueString('EventHub Namespace${baseTime}')
+var uniqueEventHubNamespaceUntrim = 'evhns-${uniqueString(resourceGroup().id)}'
```
MrMCake, line 130: @eriqua is that what you were going for in the issue? 

The obivous advantage is that the name isn't new every time the deployment runs. However, does it also ensure that no numbers are part of the `uniqueString` ?

eriqua: Numbers are not accepted in the first position but can be used after so it should be fine to have the uniqueString as a suffix.
A couple of comments:

- If the `baseTime` parameter is not used anymore we should remove it and regenerate the readme file
- The uniqueString function returns a string of 13 chars so the substring function in the next line should never be used since length(uniqueEventHubNamespaceUntrim) will never be greater than 50. Hence I'd also suggest to clean up that code

MrMCake (approved):

PR #1367, opened 2022-05-06
[Modules] SQL-Server: Added private endpoint

MrMCake:
# Description

- Added private endpoint to SQL Server (Note: slightly adjusted implementation)

| Pipeline |
| - |
|[Sql: Servers](https://github.com/Azure/ResourceModules/actions/workflows/ms.sql.servers.yml) |

# Type of Change


arm/Microsoft.Sql/servers/deploy.bicep
```diff
@@ -140,6 +143,20 @@ module server_databases 'databases/deploy.bicep' = [for (database, index) in dat
   }
 }]
 
+module server_privateEndpoints '.bicep/nested_privateEndpoint.bicep' = [for (endpoint, index) in privateEndpoints: if (!empty(privateEndpoints)) {
+  name: '${uniqueString(deployment().name, location)}-Sql-PrivateEndpoints-${index}'
+  params: {
+    privateEndpointResourceId: server.id
+    privateEndpointVnetLocation: !empty(privateEndpoints) ? reference(split(endpoint.subnetResourceId, '/subnets/')[0], '2020-06-01', 'Full').location : 'dummy'
+    service: contains(endpoint, 'service') ? endpoint.service : 'sqlServer'
```
eriqua, line 151: Should we hardcode the service name in here or use a the parameter instead? Up for debate I guess.

I see the value of hardcoding it, but leveraging the parameter would be easier to maintain in case multiple services will ever be added.

MrMCake: @eriqua I had the same thought. 
I don't really see the point of 'not' hardcoding it (as I doubt somebody will create e.g. a storage private endpoint for in an SQL server deployment) - but then again I wasn't sure if I'm missing something. Providing a 'default' seemed to be a reasonable compromise. What do you think?

eriqua: @MrMCake I'll approve since that's a more general conversation which applies also to other modules. So if we come up with a different approach we'll update all of them accordingly

eriqua (approved):
🆗

PR #1373, opened 2022-05-08
[Static Validation] Pester structural tests

MrMCake:
# Description

- Added tests for pipelines to the Pester test file
- Added documentation to configure the new settings

| Pipeline |
| - |
| [Storage: StorageAccounts](https://github.com/Azure/ResourceModules/actions/workflows/ms.storage.storageaccounts.yml)|
|[AnalysisServices: Servers](https://github.com/Azure/ResourceModules/actions/workflows/ms.analysisservices.servers.yml)|

# Type of Change


eriqua (changes requested):
I'd strongly suggest to avoid asking for another setting to customize. The majority of consumers, if not all, will use only one approach between ADO and GH. 

What about adding the logic in the test itself, to run the ADO and/or the GH tests only if respectively the .azuredevops and/or the .github folders exist? That way we could remove the additional parameters from the settings.json and corresponding wiki instructions. 

Instead, in the wiki, under the static validation, we should specify one sentence or two about the new test.

MrMCake:
> I'd strongly suggest to avoid asking for another setting to customize. The majority of consumers, if not all, will use only one approach between ADO and GH.
> 
> What about adding the logic in the test itself, to run the ADO and/or the GH tests only if respectively the .azuredevops and/or the .github folders exist? That way we could remove the additional parameters from the settings.json and corresponding wiki instructions.
> 
> Instead, in the wiki, under the static validation, we should specify one sentence or two about the new test.

done

MariusStorhaug:
Looks good, though the changes added here will only run if a workflow file exists for a module, correct? The idea with the issue that triggered this PR was to add a global pipeline/workflow to check a module compliance without having to rely on pipeline/workflow files existing for a module. I encountered many times that there exist a workflow file, but no pipeline file. That usecase would be covered here, just thinking that it should be independent of the module-workflow running. It could basically be a status check/validation pipeline triggering on PR if so. i.e. if the contribution matching the "desired module design".

MrMCake:
> Looks good, though the changes added here will only run if a workflow file exists for a module, correct? The idea with the issue that triggered this PR was to add a global pipeline/workflow to check a module compliance without having to rely on pipeline/workflow files existing for a module. I encountered many times that there exist a workflow file, but no pipeline file. That use case would be covered here, just thinking that it should be independent of the module-workflow running. It could basically be a status check/validation pipeline triggering on PR if so. i.e. if the contribution matching the "desired module design".

Almost. If only checks if a Github or Azure DevOps folder exists respectively. Quite an important detail, as customers will only use one or the other

utilities/tools/Set-ModuleReadMe.ps1
```diff
@@ -95,7 +95,6 @@ function Set-ResourceTypesSection {
         $_.type -notin $ResourceTypesToExclude -and $_
     } | Select-Object 'Type', 'ApiVersion' -Unique | Sort-Object Type -Culture en-US
 
-    $TextInfo = (Get-Culture).TextInfo
```
eriqua, line 98: Is this change intended?

arm/.global/global.module.tests.ps1
```diff
@@ -37,6 +37,37 @@ Describe 'File/folder tests' -Tag Modules {
...
+                )
+
+                $workflowsFolderName = Join-Path $RepoRoot '.github' 'workflows'
+                $workflowFileName = '{0}.yml' -f $moduleFolderName.Replace('\', '/').Replace('/', '.').Replace('Microsoft', 'ms').ToLower()
+
+                Test-Path (Join-Path $workflowsFolderName $workflowFileName) | Should -Be $true
+            }
+        }
+
+        if (Test-Path (Join-Path $repoRoot '.azuredevops')) {
```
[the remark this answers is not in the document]
MrMCake: If I recall correctly, I did test both, yes

utilities/tools/Set-ModuleReadMe.ps1
MrMCake: Yes, the variable isn't used

eriqua (approved):
🆗

PR #1387, opened 2022-05-13
[Modules] Update Eventhub Network Rules + latest APIs

### LINKED ISSUE #1356: networkAcl is not a valid property for EventHub Namespace
`networkAcls` is not a valid property for EventHub Namespace: https://github.com/Azure/ResourceModules/blob/b592eb31c9e3c0c1981bc259e4ce338163c10fa6/arm/Microsoft.EventHub/namespaces/deploy.bicep#L182-L187

Need to create a `NetworkRuleSets` child resource instead: https://docs.microsoft.com/en-us/azure/templates/microsoft.eventhub/2021-06-01-preview/namespaces/networkrulesets?tabs=bicep

ahmadabdalla:
Closes #1356 

# Description

- Created the child resources for Network Rule Sets and replaced the existing 'network ACLs' property.
- Updated module to use the latest APIs

## Pipeline references

| Pipeline |
| - |
| [EventHub: Namespaces](https://github.com/Azure/ResourceModules/actions/workflows/ms.eventhub.namespaces.yml)|

# Type of Change

# Checklist

- [ ] I did format my code

### DIFF
--- a/arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
+++ b/arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
@@ -105,9 +105,8 @@
         "systemAssignedIdentity": {
             "value": true
         },
-        "networkAcls": {
+        "networkRuleSets": {
             "value": {
-                "bypass": "AzureServices",
                 "defaultAction": "Deny",
                 "virtualNetworkRules": [
                     {
@@ -115,7 +114,7 @@
                         "action": "Allow"
                     }
                 ],
-                "ipRules": []
+                "trustedServiceAccessEnabled": false
             }
         },
         "userAssignedIdentities": {
--- a/arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
+++ b/arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
@@ -108,6 +108,12 @@
         "networkRuleSets": {
             "value": {
                 "defaultAction": "Deny",
+                "ipRules": [
+                    {
+                        "action": "Allow",
+                        "ipMask": "10.10.10.10"
+                    }
+                ],
                 "virtualNetworkRules": [
                     {
                         "subnet": {
--- a/arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
+++ b/arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
@@ -110,8 +110,10 @@
                 "defaultAction": "Deny",
                 "virtualNetworkRules": [
                     {
-                        "id": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-001/subnets/<<namePrefix>>-az-subnet-x-001",
-                        "action": "Allow"
+                        "subnet": {
+                            "id": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-001/subnets/<<namePrefix>>-az-subnet-x-001"
+                        },
+                        "ignoreMissingVnetServiceEndpoint": true
                     }
                 ],
                 "trustedServiceAccessEnabled": false
--- a/arm/Microsoft.EventHub/namespaces/deploy.bicep
+++ b/arm/Microsoft.EventHub/namespaces/deploy.bicep
@@ -43,8 +43,12 @@ param authorizationRules array = [
 @description('Optional. Configuration Details for private endpoints.For security reasons, it is recommended to use private endpoints whenever possible.')
 param privateEndpoints array = []
 
-@description('Optional. Networks ACLs, this value contains IPs to whitelist and/or Subnet information. For security reasons, it is recommended to set the DefaultAction Deny.')
-param networkAcls object = {}
+@description('Optional. Networks ACLs, this object contains IPs/Subnets to whitelist or restrict access to private endpoints only. For security reasons, it is recommended to configure this object on the Namespace.')
+param networkRuleSets object = {
+  publicNetworkAccess: 'Enabled'
+  defaultAction: 'Allow'
+  trustedServiceAccessEnabled: true
+}
 
 @description('Optional. Specifies the number of days that logs will be kept for; a value of 0 will retain data indefinitely.')
 @minValue(0)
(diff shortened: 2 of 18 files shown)


arm/Microsoft.EventHub/namespaces/deploy.bicep
```diff
@@ -43,8 +43,12 @@ param authorizationRules array = [
 @description('Optional. Configuration Details for private endpoints.For security reasons, it is recommended to use private endpoints whenever possible.')
 param privateEndpoints array = []
 
-@description('Optional. Networks ACLs, this value contains IPs to whitelist and/or Subnet information. For security reasons, it is recommended to set the DefaultAction Deny.')
-param networkAcls object = {}
+@description('Optional. Networks ACLs, this object contains IPs/Subnets to whitelist or restrict access to private endpoints only. For security reasons, it is recommended to configure this object on the Namespace.')
+param networkRuleSets object = {
```
MrMCake, line 47: I wonder if those values are considered `secure` - or if (by default) e.g. `publicNetworkAccess` should be disabled. We should probably compare this with the corresponding security baseline

arm/Microsoft.EventHub/namespaces/deploy.bicep
```diff
@@ -179,12 +187,6 @@ resource eventHubNamespace 'Microsoft.EventHub/namespaces@2021-06-01-preview' =
     zoneRedundant: zoneRedundant
     isAutoInflateEnabled: isAutoInflateEnabled
     maximumThroughputUnits: maximumThroughputUnits_var
-    networkAcls: !empty(networkAcls) ? {
```
MrMCake, line 182: Out of curiosity - is this one of those objects where we have both a property & child-resource - and deploying the namespace without the property 'wipes' the settings?

This could be tested by deploying the resource with everything and then once without the child so see if the networkAcls remain

arm/Microsoft.EventHub/namespaces/deploy.bicep
ahmadabdalla: I checked this with some resources in the library that use the same `publicNetworkAccess` property. 
- Container Registry = enabled
- Data factory = enabled
- Storage Account = enabled
- Key Vault = enabled
- Event Grid = enabled
- Log Analytics = enabled
- Machine learning = disabled
- disks = disabled

As most have it `enabled` by default. Setting this to `disabled` only allows access via private endpoints. And given there is another mechanism here is to keep this to `enabled` and still use the PaaS specific Firewall to allow service endpoints / IP whitelisting. I would suggest that if we are to enforce private endpoints only on carml resources, we roll it out via a separate issue to discuss the approach and implement this across the board. Thoughts?

arm/Microsoft.EventHub/namespaces/deploy.bicep
ahmadabdalla: Given that we are using a set of 'defaults', the `networkRuleSet` will always get deployed, hence if someone adds configuration via the parameter file, which is obviously overwritten, but if they remove the parameter and then run again, then it will go back to the default parameter value.

For the scenario above: And If we remove the 'defaults'. What happens is that lets say you configure the EH without `networkRuleSets`, and then you deploy it again and this time add `networkRuleSets` to the parameter file with specific config. The deployment goes in and applies those new configs. Removing that parameter and doing the deployment again **"Doesn't"** remove the previous settings away (unlike our friendly subnets/routetables/nsgs).

So if we do not want to override, we need to remove the 'defaults', and set an `if` condition to not trigger the `networkRuleSets`. This, however, doesn't allow enabling 'secure by default', as the default behavior from Azure is to keep it public by default whenever you deploy a vanilla EventHub namespace.

arm/Microsoft.EventHub/namespaces/deploy.bicep
eriqua: I agree that disabling network access by default would only make sense if we also enabled private endpoints by default which as per recent discussion we decided not to do.

arm/Microsoft.EventHub/namespaces/deploy.bicep
eriqua: I wouldn't consider what the portal does as secure by default. I'd here suggest not to enforce a default especially if the end result on a vanilla resource would be the same as just not enabling the child. That would be primarily to keep idempotency as per @MrMCake comment and also for consistency with the approach of not deploying child resources by default.

arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
```diff
@@ -105,17 +105,18 @@
...
-                        "id": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-001/subnets/<<namePrefix>>-az-subnet-x-001",
-                        "action": "Allow"
+                        "subnet": {
+                            "id": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-001/subnets/<<namePrefix>>-az-subnet-x-001"
+                        },
+                        "ignoreMissingVnetServiceEndpoint": true
                     }
                 ],
-                "ipRules": []
+                "trustedServiceAccessEnabled": false
```
MrMCake, line 119: Suggestion: Let's also add a simple IP rule. Like 'allow `1.1.1.1` or so.

arm/Microsoft.EventHub/namespaces/networkRuleSets/deploy.bicep
```diff
@@ -0,0 +1,69 @@
...
+
+resource namespace 'Microsoft.EventHub/namespaces@2021-11-01' existing = {
+  name: namespaceName
+}
+
+resource networkRuleSet 'Microsoft.EventHub/namespaces/networkRuleSets@2021-11-01' = {
+  name: 'default'
+  parent: namespace
+  properties: {
+    publicNetworkAccess: !(empty(ipRules)) || !(empty(virtualNetworkRules)) ? null : publicNetworkAccess
```
MrMCake, line 54:
```suggestion
    publicNetworkAccess: !empty(ipRules) || !empty(virtualNetworkRules) ? null : publicNetworkAccess
```

arm/Microsoft.EventHub/namespaces/networkRuleSets/deploy.bicep
```diff
@@ -0,0 +1,69 @@
...
+param publicNetworkAccess string = 'Enabled'
+
+@allowed([
+  'Allow'
+  'Deny'
+])
+@description('Optional. Default Action for Network Rule Set. Default is "Allow". Will be set to "Deny" if ipRules virtualNetworkRules or are being used. This would render the namespace resources inaccessible for data-plane requests if ipRules or virtualNetworkRules are not used and PublicNetworkAccess is set to "Disabled"')
+param defaultAction string = 'Allow'
+
+@description('Optional. List of IpRules. When used, defaultAction will be set to "Deny" and publicNetworkAccess will be set to "Enabled".')
```
MrMCake, line 18:
```suggestion
@description('Optional. List of IpRules. When used, the defaultAction will be set to "Deny" and publicNetworkAccess will be set to "Enabled" unless configured otherwise..')
```

arm/Microsoft.EventHub/namespaces/.parameters/parameters.json
ahmadabdalla: Done

arm/Microsoft.EventHub/namespaces/deploy.bicep
ahmadabdalla: Removed the defaults, and set a condition on the networkRuleSets module

Issue #1357, opened 2022-05-03
[Feature Request]: Add Pester test for conditional parameters description


eriqua:
### Description

Conditional parameter descriptions should include the condition under which they are required.

Suggestion: Add a Pester test to check that "`Required if *`" is included in the parameter description when the parameter description starts with "`Conditional. *`", leveraging regex.

MrMCake:
Can be closed as PR is closed

PR #1427, opened 2022-05-21
[Static Validation] Add Pester test for conditional parameters description

eriqua:
# Description

Closes issue #1357

Adding test to check if Conditional parameters' description contain a "Required if" string.

## Test references

- Tested Key Vault module by removing the `Required if` string from `keyVaultName` conditional parameter in `arm\Microsoft.KeyVault\vaults\accessPolicies\deploy.bicep` 
    - failing test https://github.com/Azure/ResourceModules/runs/6537097073?check_suite_focus=true#step:3:548
- Tested the same adding the `Required if` string back
    - successful test https://github.com/Azure/ResourceModules/runs/6537123966?check_suite_focus=true#step:3:413

# Type of Change

- [ ] Update to documentation

# Checklist

- [ ] I did format my code


arm/.global/global.module.tests.ps1
```diff
@@ -787,6 +787,36 @@ Describe 'Deployment template tests' -Tag Template {
...
 
+        It "[<moduleFolderName>] Conditional parameters' description should contain 'Required if' followed by the condition making the parameter required." -TestCases $deploymentFolderTestCases {
+            param(
+                $moduleFolderName,
+                $templateContent
+            )
+
+            if (-not $templateContent.parameters) {
+                # Skip test
+                $true | Should -Be $true
```
MrMCake, line 798: To skip the test (i.e. Pester will flag it as such), remove this dummy test.
```suggestion
```

arm/.global/global.module.tests.ps1
```diff
@@ -787,6 +787,36 @@ Describe 'Deployment template tests' -Tag Template {
...
+                $data = ($templateContent.parameters.$parameter.metadata).description
+                switch -regex ($data)
+                {
+                    '^Conditional. .*'
+                    {
+                        if ($data -notmatch '.*\. Required if .*') {
+                            $incorrectParameters += $parameter
+                        }
+                    }
+                    # Possible additional cases can be added here
```
MrMCake, line 814: What other cases do you have in mind?

arm/.global/global.module.tests.ps1
eriqua: That is being used in other tests as well. Is there a particular reason why this should be removed here? Or should we remove it from the other tests as well?

arm/.global/global.module.tests.ps1
eriqua: Not yet, but in case something comes up we may use it there. Or not. What do you suggest?

arm/.global/global.module.tests.ps1
MrMCake: It should only be there if there is no other way to 'pass' a valid test. I agree that, if it is used incorrectly in other places, we should clean it up. But I hope my rationale makes sense. I'd personally rather have Pester tell me a certain tests didn't run because it's not applicable, rather than 'pretending' it passed.

arm/.global/global.module.tests.ps1
MrMCake: I assume you're hinting at (and I know this example doesn't make sense) things like "If a parameter is required, the description must contain xzy" etc.
If so, I'd suggest to either
- Have a new test case for each future case we may want to handle or
- Rename the current test to not call out specifically for the conditional

Either are absolutely fine by me. And if you meant something completly different, too - just wanted to point it out and align :)

eriqua: Removed the comment if confusing. Let me know if there is any other specific change you're suggesting for this test.

MrMCake (approved):

Issue #1443, opened 2022-05-27
[Feature Request]: Discuss if ditching the feature to allow autogenerated unique names for resources


eriqua:
### Description

Some modules e.g. vm, evhns, storage account, allow resource names to be autogenerated if no name value is provided.

To allow deployment idempotency, the unique name uses the resourcegroup id as input of a `uniqueString` function.

The problem of this approach is that two resources of the same type cannot be deployed to the same resource group due to the collision of the autogenerated name.

Discuss if this feature provides actual value and if not remove the functionality from the modules using it, cleaning up the code and converting the corresponding name parameter as Required.

---------------------------------
Update: modules with an empty default name:

- [x] Remove autogenerated unique name Microsoft.Storage\storageAccounts\deploy.bicep
- [x] Remove autogenerated unique name Microsoft.ServiceBus\namespaces\deploy.bicep
- [x] Remove autogenerated unique name Microsoft.Network\ddosProtectionPlans\deploy.bicep
- [x] #1756
- [x] Remove autogenerated unique name Microsoft.Consumption\budgets\deploy.bicep
- [x] Remove autogenerated unique name Microsoft.Compute\proximityPlacementGroups\deploy.bicep

MrMCake:
Upvote for removing. The name generation can be implemented in the consuming deployment/template.

rahalan:
Team decides to remove this feature from a module level

PR #1447, opened 2022-05-29
[Modules] Cleared out a mountain of compiler warnings

MrMCake:
# Description

- Went through our library modules and resolved compiler warnings where possible
- Where it wasn't possible - I created an issue in the offical [Bicep thread](https://github.com/Azure/bicep/issues/784#issuecomment-1140499904)

# Pipeline reference
| Pipeline reference | 
| - |
| [Web: Connections](https://github.com/Azure/ResourceModules/actions/workflows/ms.web.connections.yml) |
| [Web: HostingEnvironments](https://github.com/Azure/ResourceModules/actions/workflows/ms.web.hostingenvironments.yml) |

# Type of Change

- [ ] Update to documentation


arm/Microsoft.RecoveryServices/vaults/replicationFabrics/replicationProtectionContainers/replicationProtectionContainerMappings/deploy.bicep
```diff
@@ -30,7 +30,7 @@ param enableDefaultTelemetry bool = true
 
 var policyResourceId = policyId != '' ? policyId : subscriptionResourceId('Microsoft.RecoveryServices/vaults/replicationPolicies', recoveryVaultName, policyName)
 var targetProtectionContainerResourceId = targetProtectionContainerId != '' ? targetProtectionContainerId : subscriptionResourceId('Microsoft.RecoveryServices/vaults/replicationFabrics/replicationProtectionContainers', recoveryVaultName, targetContainerFabricName, targetContainerName)
-var mappingName = name != '' ? name : concat(sourceProtectionContainerName, '-', split(targetProtectionContainerResourceId, '/')[10])
+var mappingName = name != '' ? name : '${sourceProtectionContainerName}-${split(targetProtectionContainerResourceId, '/')[10]}'
```
MariusStorhaug, line 33:
```suggestion
var mappingName = !empty(name) ? name : '${sourceProtectionContainerName}-${split(targetProtectionContainerResourceId, '/')[10]}'
```

arm/Microsoft.Storage/storageAccounts/deploy.bicep
```diff
@@ -152,7 +152,7 @@ var diagnosticsMetrics = [for metric in diagnosticMetricsToEnable: {
 }]
 
 var maxNameLength = 24
-var uniqueStorageNameUntrim = '${uniqueString('Storage Account${basetime}')}'
+var uniqueStorageNameUntrim = uniqueString('Storage Account${basetime}')
```
MariusStorhaug, line 155: Not sure I fully agree to the naming here. This is used in case name is not provided I take it. If this name is generated like this, it will not be idempotent. Next time it runs, a new SA will be created.
```suggestion
```

arm/Microsoft.Storage/storageAccounts/deploy.bicep
```diff
@@ -152,7 +152,7 @@ var diagnosticsMetrics = [for metric in diagnosticMetricsToEnable: {
...
 var uniqueStorageName = length(uniqueStorageNameUntrim) > maxNameLength ? substring(uniqueStorageNameUntrim, 0, maxNameLength) : uniqueStorageNameUntrim
```
MariusStorhaug, line 156: This will generate an idempotent name for the storageaccount if name is not provided.
```suggestion
var uniqueStorageName = take(uniquestring(resourceGroup().id), maxNameLength)
```

MariusStorhaug (changes requested):
Still need to check 3 files, but they seem to be more then fixing compiler warnings? Update version on some modules resulting in breaking changes on some module.

MariusStorhaug (changes requested):
Checked the last 2 files, `hostingEnvironments`. Seem to be about something out of scope for the PR. But it's done an in the PR, so lets just get it in.

arm/Microsoft.Storage/storageAccounts/deploy.bicep
MrMCake: I guess that was the idea. Anyhow - I strongly vote against changing this in this PR. We have a separate issue for this topic open here #1443.

The only change I did as part of this PR was to update the string-builder as per the Bicep warning.

MariusStorhaug: Agree

MrMCake:
> Still need to check 3 files, but they seem to be more then fixing compiler warnings? Update version on some modules resulting in breaking changes on some module.

They all relate to warnings. The changes I did in some places are actually rooted in the API updates as somebody apparanelty forgot to update the parameters accordingly (e.g. ASE). Hence there were several places where the deployment would actually not accept MULTIPLE properties (and apparently ignore them).

arm/Microsoft.Storage/storageAccounts/deploy.bicep
MariusStorhaug, line 156:
```suggestion
var uniqueStorageName = take(uniqueStorageNameUntrim, maxNameLength)
```

MariusStorhaug (changes requested):
Suggestion to simplify the code for name gen for sa

arm/Microsoft.Storage/storageAccounts/deploy.bicep
MrMCake: Suggesting to do this in a separate PR ;) Especially as we may ideally ditch the feature alltogether (see #1443)

MariusStorhaug (approved):

Issue #931, opened 2022-01-26
Investigate GitHub caching feature


MrMCake:
@renepajta found a feature that seems to allow us to cache e.g. PowerShell modules for performance optimization:

>  "GitHub-hosted runners are started as fresh environments for each job, so if your jobs regularly reuse dependencies, you can consider caching these files to help improve performance. Once the cache is created, it is available to all workflows in the same repository."

Links
- [Advanced workflow features](https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#caching-dependencies)
- [PowerShell Module Cache Marketplace action](https://github.com/marketplace/actions/powershell-module-cache)

MariusStorhaug:
As described in the documentation from the runners, they should have the modules installed. If this is to improve module loading for runners, looking into how to use the already present modules could also speed things up.

GitHub: https://github.com/actions/virtual-environments/blob/ubuntu20/20220123.1/images/linux/Ubuntu2004-Readme.md#az-powershell-modules

ADO: 
https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software

Curious.. the links on ADO point to the same GitHub page though they work differently.

MariusStorhaug:
Looked into the runner, it is storing modules under C:\Modules\az_<version>, or /usr/share/az_<version>

MariusStorhaug:
Tested `Import-Module /usr/share/az_7.1.0/Az` which triggers the import, but fails on the availability of the AzAccount module. Needs just some more investigation I think.

MariusStorhaug:
Also, any reason we are not using the AzurePowerShell built in tasks?

- [azure/powershell@v1 | GitHub](https://github.com/marketplace/actions/azure-powershell-action)
- [Azure PowerShell task | Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-powershell?view=azure-devops)

MrMCake:
> Also, any reason we are not using the AzurePowerShell built in tasks?
> 
> * [azure/powershell@v1 | GitHub](https://github.com/marketplace/actions/azure-powershell-action)
> * [Azure PowerShell task | Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-powershell?view=azure-devops)

[...]

PR #1451, opened 2022-05-30
[GitHub, CI] Update GitHub Workflows to use Azure/PowerShell action

MariusStorhaug:
# Description

Closing #931

Workflows that interact with Azure now use the Azure/PowerShell action, limiting the time it takes to set up the runner environment.

## Pipeline references

| Pipeline |
| - |
| [Compute: Galleries](https://github.com/MariusStorhaug/ResourceModules/actions/workflows/ms.compute.galleries.yml) |
| [Batch: BatchAccounts](https://github.com/MariusStorhaug/ResourceModules/actions/workflows/ms.batch.batchaccounts.yml) |
| [KeyVault: Vaults](https://github.com/MariusStorhaug/ResourceModules/actions/workflows/ms.keyvault.vaults.yml) |

# Type of Change

- [ ] Update to documentation

# Checklist

- [x] I did format my code


ahmadabdalla:
@MariusStorhaug the settings.json and the global variables are still referencing your env?

ahmadabdalla:
I'm happy with the changes. Good one @MariusStorhaug. I have question, in the scenarios where we need to bring in a custom module, like the one we use for the [environment variables YAML](https://github.com/Azure/ResourceModules/blob/main/.github/actions/templates/setEnvironmentVariables/action.yml), should we just move that 'install-module' to that specific task? or are we going to continue to support the [Set-EnvironmentOnAgent](https://github.com/Azure/ResourceModules/blob/main/utilities/pipelines/sharedScripts/Set-EnvironmentOnAgent.ps1) script?

eriqua (changes requested):
Can this be implemented for ADO as well? Or better: since we have Az PS tasks in ADO (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-powershell?view=azure-devops) maybe it's already the case that we can get rid of the Setup environment step in ADO as well and have GitHub and ADO aligned.

MrMCake:
> Can this be implemented for ADO as well? Or better: since we have Az PS tasks in ADO (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-powershell?view=azure-devops) maybe it's already the case that we can get rid of the Setup environment step in ADO as well and have GitHub and ADO aligned.

We use already the AzurePowerShell task. However the only difference is, that the task comes with the link to the ServiceConnection. We're still dependent on what is installed on the agents. The same is true for the Agents in GitHub too by the way.

.github/actions/templates/publishModule/action.yml
```diff
@@ -64,26 +64,6 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: 'Setup agent'
```
MrMCake, line 67: Why removing this?

The script only checks if the modules are installed and if they are prints out their version. Or at least it is supposed to. For example:

```
  VERBOSE: Try installing:
  VERBOSE: - 1. [Az.Accounts]
  VERBOSE: - 2. [Az.Resources]
  VERBOSE: - 3. [Az.CognitiveServices]
  VERBOSE: - 4. [Az.Compute]
  VERBOSE: - 5. [Az.KeyVault]
  VERBOSE: - 6. [Az.Monitor]
  VERBOSE: - 7. [Az.OperationalInsights]
  VERBOSE: - 8. [Az.RecoveryServices]
  VERBOSE: Install-CustomModule start
  VERBOSE: =====================
  VERBOSE: HANDLING MODULE [1/8] [Az.Accounts] 
  VERBOSE: =====================
  VERBOSE: Install module [Az.Accounts] with version [2.8.0]
  VERBOSE: Module [Az.Accounts] is installed with version [2.8.0]
```

If the modules are in fact installed and the script cannot find them (though I have seen it working in the past), I can only assume there is a configuration missing.

eriqua:
> > Can this be implemented for ADO as well? Or better: since we have Az PS tasks in ADO (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-powershell?view=azure-devops) maybe it's already the case that we can get rid of the Setup environment step in ADO as well and have GitHub and ADO aligned.
> 
> We use already the AzurePowerShell task. However the only difference is, that the task comes with the link to the ServiceConnection. We're still dependent on what is installed on the agents. The same is true for the Agents in GitHub too by the way.

@MrMCake I understand the suggestion would be to keep the Setup agent step besides the AzurePowerShell task in ADO or action in GH. This would be since that is expected not to be time consuming for the pipeline to be executed in case Az PS modules are already there. My point is in both cases, both if we decide to remove and if we decide to keep it, that should be consistent between ADO and GH. So either we remove that from both or we keep it in both.

.github/actions/templates/publishModule/action.yml
eriqua: I strongly support any update simplifying the code and reducing the pipeline duration. I'd have 2 points:

1. Since we're mentioning self-hosted, the Azure PowerShell action currently only supports ubuntu and windows runners. Self hosted runners are not supported (ref https://github.com/marketplace/actions/azure-powershell-action). We probably need to mention that in the documentation. I think we cannot expect the CI environment to be working as is with self-hosted agents/runners and probably cannot commit to that.
3. We need to agree if we want to consider the module versions in our scripts, meaning if we're good with using whichever version used by the Az PS tasks or if we expect the need to be able to force a specific module version to be used. In the latter case we probably need to keep the Setup agents step.

MariusStorhaug: I don't understand why we ensure that these are installed, when they are there by default (and loaded) when using the AzurePowerShell action. To me it's bloating code which give no clear added benefit.

MariusStorhaug: As the runner states, it's using the following software: https://github.com/actions/virtual-environments/blob/ubuntu20/20220515.1/images/linux/Ubuntu2004-Readme.md#az-powershell-modules

Here it states it has 7.3.2 installed, and using the AzurePowerShell action it would load these module. According to https://www.powershellgallery.com/packages/Az/7.3.2 its installing [Az.Accounts](https://www.powershellgallery.com/packages/Az.Accounts/) (>= 2.7.4). So its there, but its not using the absolute newest version. Do we have requirements where we fully need to control the version of the RP level module or is this a 'nice to have'? It makes sense (and has not been removed) where we load other modules which are non Az modules.

MariusStorhaug: To me it's more building a list of requirements for runners/agents and make sure that runners have the necessary software in place up-front and not have that run every time the CI is running, costing execution time. The task we are trying to close is regarding improving execution time. Keeping the code and the execution of this script (which is unnecessary given the pre-installed software on the runners) means that we don't really needed the issue in the first place?

MrMCake:
> > > Can this be implemented for ADO as well? Or better: since we have Az PS tasks in ADO (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-powershell?view=azure-devops) maybe it's already the case that we can get rid of the Setup environment step in ADO as well and have GitHub and ADO aligned.
> > 
> > 
> > We use already the AzurePowerShell task. However the only difference is, that the task comes with the link to the ServiceConnection. We're still dependent on what is installed on the agents. The same is true for the Agents in GitHub too by the way.
> 
> @MrMCake I understand the suggestion would be to keep the Setup agent step besides the AzurePowerShell task in ADO or action in GH. This would be since that is expected not to be time consuming for the pipeline to be executed in case Az PS modules are already there. My point is in both cases, both if we decide to remove and if we decide to keep it, that should be consistent between ADO and GH. So either we remove that from both or we keep it in both.

Hey @eriqua, gotcha - and agree. Just wanted to make the point about the type of task. :)

.github/actions/templates/publishModule/action.yml
MrMCake: In theory, the task should only take a couple seconds as it should have found the already installed modules and just continue. However, for some reason, the modules were not found and hence it always spent a minute to install the 'seamingly' missing modules. I agree, that we should not re-install modules that are already there, as it just wastes time - hence it might 
The linked issue assumes that there are actually modules missing - and if it is required to install them - then it would be intersting to leverage that 'caching' functionality to not do it over and over again.

MrMCake: I did some further [testing](https://github.com/Azure/ResourceModules/runs/6706128606?check_suite_focus=true) and found some interesting insights.

In general, most time is consumed not by the PowerShell modules, but by the 'az --version', 'bicep --version' etc. (in total, 20 seconds)
Regarding only the PowerShell modules, if the modules are found on the machine, it takes 11 seconds to find 3 of them and print the output. If they are installed, it takes 30 seconds.

What's more interesting though is the fact that the overall azure/powershell task still takes more than a minute to complete. I can only assume that the task does some other stuff in the background.

I was able to achieve the [best outcome](https://github.com/Azure/ResourceModules/runs/6706212474?check_suite_focus=true) with all outputs enabled by using the following setup:
- Not use the Azure/powershell task
- Add the path to th installed modules, used by the Azure/PowerShell task myself using `$Env:PSModulePath += ":/usr/share/az_7.3.2"`

MariusStorhaug: Another issue was added for this work. #1466

.github/actions/templates/validateModuleDeployment/action.yml
```diff
@@ -110,6 +110,7 @@ runs:
 
         Write-Output "::endgroup::"
 
+
```
MariusStorhaug, line 113:
```suggestion
```

MrMCake (approved):

MrMCake (approved):

Issue #644, opened 2021-11-29
Modules: Enable removal of remaining resources that need  a custom script


MrMCake:
Now that we switched the removal to deployment names there are less resources we cannot remove. However there are still a few that we need to tackle, such as:

- [x] Diagnostic Settings
- [x] Managed Services Registration Definitions
- [ ] Security Center
- [x] Locks

eriqua:
Diagnostic Settings are included

MrMCake:
@eriqua The PRs for Locks & RegistrationDefinition are in now. 

Regarding the Security Center I wonder if we shoud implement the removal. The SecurityCenter is somewhat 'global' and we could accidently remove legit settings. What do you think?

eriqua:
I think is fine if we keep the SecurityCenter removal disabled by default, but we should probably anyway have the removal logic working in case the removal flag is enabled. Also we probably need to specify the reasons of the exception somewhere in the wiki (e.g. the removal section in the deployment validation page)

eriqua:
See new issue #1524 for azure security center

PR #1479, opened 2022-06-04
[Pipelines] Added removel for registrationDefinitions

MrMCake:
# Description

- Added removel for registrationDefinitions
- Enabled removal in pipeline

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation


utilities/pipelines/resourceRemoval/helper/Invoke-ResourceRemoval.ps1
```diff
@@ -77,6 +77,16 @@ function Invoke-ResourceRemoval {
             $roleAssignmentsOnScope | Where-Object { $_.RoleAssignmentId -eq $ResourceId } | Remove-AzRoleAssignment
             break
         }
+        'Microsoft.ManagedServices/registrationAssignments' {
```
eriqua, line 80: Are we sure all this custom logic is needed and it is not enough to just enable the removal by default?

MrMCake: It's a result of the corresponding issue #644 - hence I expected it to be required. I'll go ahead and run it once and force it to try and remove the definition

MrMCake: True story. Just testedit in main and it 'seems' to work. Undid the changes now and re-run from this branch.

Issue #776, opened 2021-12-09
Set default values to a secure value - Cognitive Services


elbatane:
All default values should comply with a security baseline, e.g. [NIST 800](https://docs.microsoft.com/en-us/azure/governance/policy/samples/nist-sp-800-53-r5)

The [build-in policies of Azure](https://github.com/Azure/azure-policy) can be used as a reference.

The task would be to scan over each of the following policies and make sure, that the module is per default complying to them.

The following policies are the NIST 800 ones:

- [x] \built-in-policies\policyDefinitions\Cognitive Services\CognitiveServices_CustomerManagedKey_Audit.json
- [x] \built-in-policies\policyDefinitions\Cognitive Services\CognitiveServices_DisableLocalAuth_Audit.json
- [x] \built-in-policies\policyDefinitions\Cognitive Services\CognitiveServices_DisablePublicNetworkAccess_Audit.json
- [x] \built-in-policies\policyDefinitions\Cognitive Services\CognitiveServices_EnablePrivateEndpoints_Audit.json
- [x] \built-in-policies\policyDefinitions\Cognitive Services\CognitiveServices_NetworkAcls_Audit.json

For Private Endpoint compliance, add the privateEndpoints parameter to the max parameters.json file and update the privateEndpoints parameter description in the bicep.json to indicate private endpoints should be used. For example:

```json
@description('Optional. Configuration Details for private endpoints. Cognitive Services accounts should use private endpoints.')
param privateEndpoints array = []
```

MrMCake:
CMK implementation is current on hold until [this raised issue[(https://github.com/Azure/bicep/issues/7066) is sorted. I started implementing it in branch 'users/alsehr/776_cogSerCMK'.

MrMCake:
Also, note that the feature seems still in Preview aus you have to fill out a form: https://docs.microsoft.com/en-us/azure/cognitive-services/language-service/concepts/encryption-data-at-rest#customer-managed-keys-for-language-services

PR #1453, opened 2022-05-31
[Modules] Service bus private networking ACL

ChrisSidebotham:
# Description

Private Networking for Azure Service Bus

Adding of the networkRuleSets Module to allow network configurations to be dynamically created for Azure Service Bus based on SKU and Private Endpoint 

## Pipeline references

| Pipeline |
| - |
| [ServiceBus: Namespaces](https://github.com/ChrisSidebotham/ResourceModules/actions/workflows/ms.servicebus.namespaces.yml) |

https://github.com/ChrisSidebotham/ResourceModules/actions/runs/2410217790

# Type of Change

Please delete options that are not relevant.

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Update to documentation

# Checklist

- [x] I'm sure there are no other open Pull Requests for the same update/change
- [x] My corresponding pipelines / checks run clean and green without any errors or warnings
- [x] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (readme)
- [x] I did format my code

Issue #819, opened 2021-12-16
Private endpoints and network Acls - to be discussed


elanzel:
When implementing private endpoints a network Acls with the parameter "defultAction" set to "Deny" needs to be put in place at the same time:

"privateEndpoints": {
            "value": [
                {
                    "subnetResourceId": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-sxx-az-vnet-x-001/subnets/sxx-az-subnet-x-005-privateEndpoints",
                    "service": "vault"
                }
            ]
        },
        "networkAcls": {
            "value": {
                "bypass": "AzureServices",
                "defaultAction": "Deny",
                "virtualNetworkRules": [],
                "ipRules": []
            }
        }

Should this "dependency" be forced by the code? Or should we show this in the parameters file and in the readme? Or are there other alternatives?

rahalan:
1. team decides to provide this setting by code, unless a networkAcl is provided
2. this potentially is true for other modules (probably storage account)

rahalan:
=> if it applies to more than KeyVault, please create separate issues and link to this one

eriqua:
Decision out of today's discussion:

- if networkAcls are explicitly passed in parameters keep what is passed which has priority, besides if pe is enabled or not.
- Otherwise:
   - if pe is enabled -> disable network public access by default
   - if pe is not enabled -> do nothing explicitly on networkAcls but keep the default

eriqua:
Updated issue description
cc: @ahmadabdalla

AlexanderSehr:
@elanzel could it be that this issue is already implemented? We definitely implemented the NetworkAcls + modified the public network access based on whether Private Endpoints are provided or not.

AlexanderSehr:
Hey @eriqua & @elanzel, any take on this? I still agree with my past self 😄

AlexanderSehr:
I think we did 😄 or at least all the implementations looked suspiciously simiar 😉

PR #1483, opened 2022-06-05
[Modules] Congnitive Services: Set secure defaults

MrMCake:
# Description

- Set secure defaults based on #776 

| Pipeline |
| - |

# Type of Change

- [ ] Update to documentation


arm/Microsoft.CognitiveServices/accounts/deploy.bicep
```diff
@@ -73,20 +73,20 @@ param diagnosticEventHubName string = ''
 @description('Conditional. Subdomain name used for token-based authentication. Required if \'networkAcls\' are set.')
 param customSubDomainName string = ''
 
-@description('Optional. Whether or not public endpoint access is allowed for this account.')
+@description('Optional. Whether or not public endpoint access is allowed for this account. Should be disabled for security reasons.')
```
eriqua, line 76: As we cannot enable private endpoints by default on resources I don't think we can disable public network access by default. Without private endpoints this would mean locking access to resources completely.

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
```diff
@@ -209,7 +203,11 @@ resource cognitiveServices 'Microsoft.CognitiveServices/accounts@2021-10-01' = {
   }
   properties: {
     customSubDomainName: !empty(customSubDomainName) ? customSubDomainName : null
-    networkAcls: !empty(networkAcls) ? networkAcls_var : null
+    networkAcls: !empty(networkAcls) ? {
+      defaultAction: contains(networkAcls, 'defaultAction') ? networkAcls.defaultAction : 'Deny'
+      virtualNetworkRules: contains(networkAcls, 'virtualNetworkRules') ? networkAcls.virtualNetworkRules : []
+      ipRules: contains(networkAcls, 'ipRules') ? networkAcls.ipRules : []
+    } : null
```
eriqua, line 210: Same as disabling public network access I'd keep the default behavior here
```suggestion
    networkAcls: !empty(networkAcls) ? {
      defaultAction: contains(networkAcls, 'defaultAction') ? networkAcls.defaultAction : null
      virtualNetworkRules: contains(networkAcls, 'virtualNetworkRules') ? networkAcls.virtualNetworkRules : []
      ipRules: contains(networkAcls, 'ipRules') ? networkAcls.ipRules : []
    } : null
```

arm/Microsoft.CognitiveServices/accounts/readme.md
```diff
@@ -347,7 +399,7 @@ userAssignedIdentities: {
 
 ## Considerations
 
-- Not all combinations of parameters `kind` and `sku` are valid and they may vary in different Azure Regions. Please use PowerShell CmdLet `Get-AzCognitiveServicesAccountSku` or another methods to determine valid values in your region.
+- Not all combinations of parameters `kind` and `SKU` are valid and they may vary in different Azure Regions. Please use PowerShell CMDLet `Get-AzCognitiveServicesAccountSku` or another methods to determine valid values in your region.
```
eriqua, line 402:
```suggestion
- Not all combinations of parameters `kind` and `SKU` are valid and they may vary in different Azure Regions. Please use PowerShell cmdlet `Get-AzCognitiveServicesAccountSku` or another methods to determine valid values in your region.
```

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
MrMCake: Agreed. Just following the security policy here. Any counterproposal? Should I enable it again and just keep the 'Hey you should disable this' description? :D

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
MrMCake: Same comment as above. What takes precidence: Security recommendations, or module usability? I honestly can't tell.

eriqua: I think this is dependent on a wider discussion as per issue #819, similarly to PR #1453. I've put that back to triage and applied the needs further discussion label so that we can align on a common approach for all modules. In general I think we should if possible consider all the 3 components of the CIA triad when dealing with security: Confidentiality, Integrity and Availability. By locking resource access we're not taking Availability into account, which is still part of the Security topic.

MrMCake: Agreed. The only thing we make sure is, that if somebody e.g. deploys the module with a private endpoint, they don't forget to lock the service down

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
MrMCake: Updated

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
```diff
@@ -209,8 +204,12 @@ resource cognitiveServices 'Microsoft.CognitiveServices/accounts@2021-10-01' = {
   }
   properties: {
     customSubDomainName: !empty(customSubDomainName) ? customSubDomainName : null
-    networkAcls: !empty(networkAcls) ? networkAcls_var : null
-    publicNetworkAccess: publicNetworkAccess
+    networkAcls: !empty(networkAcls) ? {
+      defaultAction: contains(networkAcls, 'defaultAction') ? networkAcls.defaultAction : null
```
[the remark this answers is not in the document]
MrMCake: Removed. The deployment of the private endpoint will disable network access by default

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
```diff
@@ -70,23 +70,24 @@ param diagnosticEventHubAuthorizationRuleId string = ''
 @description('Optional. Name of the diagnostic event hub within the namespace to which logs are streamed. Without this, an event hub is created for each log category.')
 param diagnosticEventHubName string = ''
 
-@description('Conditional. Subdomain name used for token-based authentication. Required if \'networkAcls\' are set.')
+@description('Conditional. Subdomain name used for token-based authentication. Required if \'networkAcls\' are set (including private endpoints).')
```
eriqua, line 73: Not sure I understand what you mean with `(including private endpoints)`

.azuredevops/modulePipelines/ms.cognitiveservices.accounts.yml
```diff
@@ -47,6 +47,7 @@ stages:
           deploymentBlocks:
             - path: $(modulePath)/.parameters/parameters.json
             - path: $(modulePath)/.parameters/speech.parameters.json
+            - path: $(modulePath)/.parameters/encr.parameters.json
```
eriqua, line 50: Should we add a min.parameters?

arm/Microsoft.CognitiveServices/accounts/.parameters/parameters.json
```diff
@@ -24,6 +24,15 @@
                 }
             ]
         },
+        "networkAcls": {
+            "value": {
+                "defaultAction": "deny",
+                "virtualNetworkRules": []
+            }
+        },
```
eriqua, line 32: Why this?

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
MrMCake: Private Endpoints are not part of the NetworkAcls, but network-related and also require a subdomain name. 

I guess I could also say `Required if \'networkAcls\' or \'privateEndpoints\' are set.`?

arm/Microsoft.CognitiveServices/accounts/.parameters/parameters.json
MrMCake: Came out of a discussion with Marius to test if it works. I.e. can I set network rules without specifying the entire object.

eriqua: I see, I'd probably use different tests to test the networking and instead have the private endpoint parameters in the main parameter.json file. What do you think?

arm/Microsoft.CognitiveServices/accounts/deploy.bicep
MrMCake: Updated

.azuredevops/modulePipelines/ms.cognitiveservices.accounts.yml
MrMCake: Added

arm/Microsoft.CognitiveServices/accounts/.parameters/parameters.json
MrMCake: The reason it is in a different file is that the Face API (currently used in the `parameters.json`) does not support private endpoints. Should I remove that test and only deploy the speech service?

eriqua: Then how about renaming parametrs.json to face.parameters.json?
Back to the original comment my point is more generally about the fact that since we reference "examples" in the readme I'd expect them to make sense and not only test the feasibility of each combination. However, I wouldn't block this PR for this comment.

MrMCake: ok, then let me test adding a network rule. I guess that would be a valid scenario.

MrMCake: I can rename it to `face.` by the way - though then we don't have an obvious parameter file for tests like 'diagnostic settings' anymore :D

MrMCake: Updated and re-running tests

MrMCake: done & green

eriqua: Wouldn't you see that similar to what we have for web sites and vms? we don't have a parameters.json there but the naming is following the specific use case. If we wanted the same here, we should have:

- face.parameters.json
- speech.parameters.json
- speech.min.parameters.json
- speech.encr.parameters.json

It's not strictly related to this PR so again I don't want to block it, but we should agree on some consistency for the parameter file naming

eriqua: Opened issue #1564 to continue the conversation there and to not block this PR

eriqua (approved):
🆗

PR #1502, opened 2022-06-07
[CI environment] Updated the triggerpath from ..pipelineTemplates/module.*  to pipelineTemplates/jobs.*

akata72:
# Description

There seems to be no files with the /module.* pattern in the pipelineTemplates folder, based on the history it looks like this is the old names of the YML templates currently in the folder. So I guess the trigger path should also be changed to reflect this?? 

Changing this would start triggering all pipelines when the templates are changed (this currently doesn't happen, at least in my ADO pipelines). 

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation

# Checklist

- [x] I did format my code


MrMCake:
> Yes. It does honor these wildcards. Just tested it. When changing any of the templates, the module pipeline is triggered. It means all module pipelines will trigger when any of the templates are updated. Is that the original goal of having this trigger?

It is. It is the only way we can make sure the pipelines still work as intended. It would be better if there would be a way to be more ... specific - for example only run the Static Tests in case their template changed - but unfortunatelty it's all or nothing.

.azuredevops/modulePipelines/ms.aad.domainservices.yml
```diff
@@ -20,7 +20,7 @@ trigger:
   paths:
     include:
       - '/.azuredevops/modulePipelines/ms.aad.domainservices.yml'
-      - '/.azuredevops/pipelineTemplates/module.*.yml'
+      - '/.azuredevops/pipelineTemplates/jobs.*.yml'
```
MrMCake, line 23: The only thing I'm wondering is, if it would make more sense to just use `'/.azuredevops/pipelineTemplates/*.yml'`.

On the upside, any future potential change would be covered. 

However, there may be a downside to it if somebody adds a template that is not related to the module pipelines in that folder. For example a template for the dependencies. Any thoughts?

@eriqua & @MariusStorhaug would also be curious to hear your thoughts in the matter. 

It is in any case a small change - I just question what would be the most optimal solution.

eriqua: In GH workflows we have `'.github/actions/templates/**'` so for consistency I'd vote for covering the whole folder here as well with `'/.azuredevops/pipelineTemplates/*.yml'`

Anyway, in case not related pipelines are added to those 2 folders in the future, we will notice and manage the issue accordingly (e.g. creating subfolders for modulePipelines and dependencies)

MrMCake: I'd suggest the opposite - use `.yml` in GH too. There was a time when we wanted to add ReadMEs and stuff - in case we every do, let's be precise now

.azuredevops/modulePipelines/ms.aad.domainservices.yml
```diff
@@ -20,7 +20,7 @@ trigger:
   paths:
```
MrMCake, line 20: As it was just merged, let's also include the new `Microsoft.DataProtection.Backupvaults` module

.azuredevops/modulePipelines/ms.aad.domainservices.yml
MariusStorhaug: I am fine with the suggestion from @MrMCake and @eriqua: keeping the `*.yml` -> But then we must update that for the GH side as well. Feel free to add that in the PR.

akata72: On the GH side there is also a folder with a PS script that will not be included if we use *.yml as the wildcard, so that might advocate the use of ** as suggested by @MariusStorhaug ..  any READMEs are excluded for both GH and ADO anyway.  Or we could use multiple wildcards of course. ?

MrMCake: @akata72 you're right. `**` it is. Consisteny for the win

.azuredevops/modulePipelines/ms.aad.domainservices.yml
akata72: Done.

.azuredevops/modulePipelines/ms.aad.domainservices.yml
akata72: I tested using ** or * for this pattern and as expected it will trigger the pipelines on any change in the folder+ . But the problem I noticed was that it also triggers when adding i.e a README.md file (that I would expect to be excluded).  So to solve the initial issue I had with pipelines not triggering with the existing path, maybe we could just use *.yml .. and create a different issue on the fact that the exclude pattern for .md files does not seem to always work.

eriqua: In a folder called `pipelineTemplates` we shouldn't have file extensions different than `.yml` or `.md`. PS scripts should be in the utilities folder together with the others. 
We need to restructure this a bit for GH moving PS scripts from `.github\actions` to `utilities` to have both triggers (gh + ado) on the `*.yml` extension. [Ref #1527 ]

For the scope of this PR I think we're good to go. @MrMCake @MariusStorhaug can we merge this?

MariusStorhaug (approved):
LGTM

Issue #1493, opened 2022-06-06
[Feature Request]: Disuss the default 'passing locks between modules' behavior


MrMCake:
### Description

Building on top of
- #239 and
- #1446

It turned out there is a missing alignment how to treat locks across all module references.

Up until now we agreed
- Children should not receive the lock from their parent via our template (as this is done automatically in Azure) unless specified otherwise in the parameter file
- Cross-module referenced modules should receive the lock from the invoking module (e.g. PIP for VM) unless specified otherwise in the parameter file

However, regarding the 2nd point, it is unclear how to treat 'Extension' resources such as Private Endpoints. Technically, they are cross-referenced, but that doesn't mean it necessarily makes sense to lock them automatically with the invoking module/

rahalan:
Team agrees on the suggestion

PR #1515, opened 2022-06-08
[CI Environment] Fixed lock removal + added error handling in case no deployment resource if found

MrMCake:
# Description

- The removal of locks takes a moment to propagate, hence I added a `Start-Sleep` in a dedicated resource type handler
- Also I noticed the pipeline fails if the deployment name results in to removalable resources. Handled a corresponding handler

| Pipeline |
| - |
| [Web: StaticSites](https://github.com/Azure/ResourceModules/actions/workflows/ms.web.staticsites.yml) |
| [KeyVault: Vaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.keyvault.vaults.yml) |
| [KeyVault: Vaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.keyvault.vaults.yml) |
| [Web: Sites](https://github.com/Azure/ResourceModules/actions/workflows/ms.web.sites.yml) |
| [MachineLearningServices: Workspaces](https://github.com/Azure/ResourceModules/actions/workflows/ms.machinelearningservices.workspaces.yml) |
| [Storage: StorageAccounts](https://github.com/Azure/ResourceModules/actions/workflows/ms.storage.storageaccounts.yml) |
| [AppConfiguration: ConfigurationStores](https://github.com/Azure/ResourceModules/actions/workflows/ms.appconfiguration.configurationstores.yml) |

# Type of Change

- [ ] Update to documentation


utilities/pipelines/resourceRemoval/helper/Invoke-ResourceRemoval.ps1
```diff
@@ -39,6 +39,15 @@ function Invoke-ResourceRemoval {
             }
             break
         }
+        'Microsoft.Authorization/locks' {
```
eriqua, line 42: Would you keep this besides the decision out of issue #1493?

MrMCake: Not sure I fully crasp the connection. The reason I added the 'case' was the sleep that I did not want to introduce to the general logic (to not waste time).
I even had a while-loop before, but it turned out, the lock is 'considered' removed directly - hence only the sleep seemed to be a viable alternative. Can't say I like it and would have preferred the loop, but it unfortunately didn't do the trick.

Issue #1363, opened 2022-05-05
[Feature Request]: Discuss rolling out the <<resourceGroupName>> token across all pipeline parameter files


ahmadabdalla:
### Description

We currently support a token called `<<resourceGroupName>` which is replaced by the `resourceGroupName` variable in the global variables YAML. 

If someone wants to change the resource group name to not be `validation-rg`, then they'd need to update all pipelines to do that as it is hardcoded in the parameter files. 

Should we roll this token out to all parameter files to allow changing the resource group name without impacting the parameter files?

MariusStorhaug:
This is as requested by the issue: #953
Would be great to have _all_ combined into one, available as tokens or variables in the pipeline/workflows. So yes, you have my support! (smiley)

ahmadabdalla:
@MariusStorhaug `global.variables.yml` will be `settings.yml` as per your suggestion.

ahmadabdalla:
@MariusStorhaug on your point:
 _Ideally, it should be fully dynamic so that we have automation to just make all the variables in the global.variable.yml file available as tokens._

In the case of GitHub, we are taking the whole file and injecting it as environment variables. Is the above alluding to taking each environment variable, appending the token prefix and suffix and then searching for it in the file? like this:

<<envVariable1>>

MariusStorhaug:
@ahmadabdalla, then I think I was unaware of how the tokens were made available in the piepline/workflows. My bad. You can disregard the point made on making them automatically available as env vars.

MrMCake:
While I think this could be enabled at any time - I wonder if it is still relevant once the new dependency approach #1791 is merged for all modules. There, the resource group name is usually passed as a reference in to the test.

ahmadabdalla:
@MrMCake with a recent feedback I got from a carml customer is that changing the RG name on the settings.yml doesn't reflect in the parameter files. Would there still be any use cases for the dependency pipelines to continue to be supported? If yes, then maybe we can roll this out?

MrMCake:
> If yes, then maybe we can roll this out?
Currently, there would not be a use case for the dependency pipeline to be supported in the future.

ahmadabdalla:
@MrMCake then we can close this ticket if there is no longer a need for it

eriqua:
Moving back to needs triage to agree on this closure. I'd anyway suggest to open an issue to remove the resource group parameter from settings.yml and all related utility referencies.

rahalan:
[...]

PR #1516, opened 2022-06-09
[CI Environment] Token Mechanism Uplift (support tokens as GitHub Secret + Migrate Settings.Json to Settings YAML)

### LINKED ISSUE #1440: [Feature Request] BREAKING change: Protect `namePrefix` value to further mitigate pipeline failures due to existing globally unique names
### Context

Managing the namePrefix token through the settings.json mitigated the risk of losing ownership on the globally unique resource names such as keyvaults. 
It turns out that it is not enough since those names keep on being used by others, most likely carml users who forget to change the namePrefix value before running the pipelines. That is causing related pipelines to fail in the CARML testing environment due to resource names being already taken.
We currently need to change those names and further regenerate readmes containing 1:1 module samples to solve corresponding pipelines failures.
This is expected to get worse with the increase of the CARML usage and with the introduction of the new dependencies approach, since also dependencies will be removed and the number of failing pipelines is likely to increase.

### Suggestion

- Create a new GH secret/ADO variable to host the namePrefix
- Keep the possibility to manage the token via the settings.json
- Implement a logic to give priority to the settings.json token if set, default to the secret otherwise
- In CARML comment out the namePrefix token in the settings.json and remove our value for the CI environment to leverage the env variable.
- Update the test module locally scripts accordingly
- Update documentation

cc: @ahmadabdalla, @MrMCake

ahmadabdalla:
Closes #1440 

# Description

- Migrated from global.variables.yml and settings.json into a single `settings.yml`
- Support for a GitHub secret / ADO variable named `TOKEN_NAMEPREFIX` which takes precedence over the `localToken_namePrefix` that exists in the settings.yml
- Updated documentation

# Details

- Pester tests changed to now receive an input of the token as a configuration instead of the pester test file having to process the settings and extract content from it.
- unified the experience for token replacement across multiple scripts and pipelines
- GitHub Module Workflows contain a new Environment Variable called `TOKEN_NAMEPREFIX`, this is not used if the secret is not specified
- Updated documentation especially referencing the need for the `powershell-yaml` module in the scenario 2 - contribution documentation.

## Experience for new TOKEN

If not defined as a secret in GHA / variable in ADO:

image

If defined as a secret in GHA/ADO variable:

image
image

| Pipeline |
| - |
| [Compute: Disks](https://github.com/ahmadabdalla/ResourceModules/actions/workflows/ms.compute.disks.yml)|
|[Authorization: RoleDefinitions](https://github.com/ahmadabdalla/ResourceModules/actions/workflows/ms.authorization.roledefinitions.yml)|
|[Network: ApplicationSecurityGroups](https://github.com/ahmadabdalla/ResourceModules/actions/workflows/ms.network.applicationsecuritygroups.yml)|

# Type of Change

- [ ] Update to documentation

# Checklist

- [ ] I did format my code


global.variables.yml
```diff
@@ -3,6 +3,26 @@ variables:
...
+  # See: https://github.com/Azure/ResourceModules/wiki/The%20library%20-%20Module%20design#telemetry
+  enableDefaultTelemetry: false
+
+  ######################################
+  # Local tokens settings
+  ######################################
+
+  # the 'localToken_' prefix will be removed from the key name when the pipelines run.
+  # e.g. if you have a token in your parameter file as <<customKey>>, then the token defined in this file looks like "localToken_customKey": 'value'
+  localToken_namePrefix: 'zu3bx' # A 3-5 character length string, included in the resources names, is overridden if provided as GitHub Secret / ADO Variable
```
eriqua, line 16: In the final version to be merged what should we have here? Is it supported to have it empty?

ahmadabdalla: The token can become default as a placeholder and no longer 'carml' as we'll move to the secret version (I presume). Removing the token shouldn't impact. However not having the local one and the secret one will result in parameters not getting the namePrefix token swapped

ahmadabdalla:
@MrMCake  @eriqua @MariusStorhaug ready for first round of review

settings.yml
```diff
@@ -3,6 +3,26 @@ variables:
...
+  # See: https://github.com/Azure/ResourceModules/wiki/The%20library%20-%20Module%20design#telemetry
+  enableDefaultTelemetry: false
+
+  ######################################
+  # Local tokens settings
+  ######################################
+
+  # the 'localToken_' prefix will be removed from the key name when the pipelines run.
+  # e.g. if you have a token in your parameter file as <<customKey>>, then the token defined in this file looks like "localToken_customKey": 'value'
+  localToken_namePrefix: 'zu3bx' # A 3-5 character length string, included in the resources names, is overridden if provided as GitHub Secret / ADO Variable
```
ahmadabdalla, line 16: placeholder comment to be updated to revert back the prefix here and in the bicep publishing name

.azuredevops/platformPipelines/platform.dependencies.yml
```diff
@@ -25,7 +25,7 @@ trigger: none
 #       - 'utilities/pipelines/dependencies/**'
 
 variables:
-  - template: '../../global.variables.yml'
```
MrMCake, line 28: Did you run the pipeline end to end?

.github/workflows/platform.dependencies.yml
```diff
@@ -23,6 +23,7 @@ on:
   #     - 'utilities/pipelines/dependencies/**'
 
 env:
```
MrMCake, line 25: Did you run the pipeline end to end?

settings.yml
```diff
@@ -30,7 +50,7 @@ variables:
   ######################################
 
   bicepRegistryDoPublish: true # Set to true, if you would like to publish module templates to a bicep registry
-  bicepRegistryName: adpsxxazacrx001 # The name of the bicep registry (ACR) to publish to. If it does not exist, it will be created.
+  bicepRegistryName: adpzu3bxazacrx001 # The name of the bicep registry (ACR) to publish to. If it does not exist, it will be created.
```
MrMCake, line 53: Please also revert the other settings

docs/wiki/The CI environment - Token replacement.md
```diff
@@ -39,37 +39,38 @@ These are tokens constructed from environment variables, which are defined in th
...
-        "value": "<<tokenName>>-keyVault"
+        "value": "<<tokenA>>-keyVault-<<tokenB>>"
     }
 }
 ```
 
-Once the Key Vault is deployed, you'll notice that the Key Vault name in Azure will be `tokenValue-keyVault`
+Once the Key Vault is deployed, you'll notice that the Key Vault name in Azure will be `foo-keyVault-bar`
+
+The token prefix `'<<'` and suffix `'>>'` in the above example are also configurable in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml) file.
```
MrMCake, line 64:
```suggestion
The token prefix `'<<'` and suffix `'>>'` in the above example are also configurable in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml) file and are used to identify the tokens in the files.
```

docs/wiki/The CI environment - Token replacement.md
```diff
@@ -39,37 +39,38 @@ These are tokens constructed from environment variables, which are defined in th
...
+
+The solution comes with one predefined local token `namePrefix`. This token is leveraged in most of the parameter files for test deployments. It allows using a consistent naming prefix that is applied to all resources being tested. There are two ways this token can be set and one will take precedence over the other:
```
MrMCake, line 66:
```suggestion
The solution comes with one predefined local token `namePrefix`. This token is leveraged in most of the parameter & test files for deployments. It allows using a consistent naming prefix that is applied to all resources being tested. There are two ways this token can be set and one will take precedence over the other:
```

.azuredevops/modulePipelines/ms.aad.domainservices.yml
```diff
@@ -28,7 +28,7 @@ trigger:
       - '/**/*.md'
 
 variables:
-  - template: '../../global.variables.yml'
+  - template: '../../settings.yml'
```
eriqua, line 31: Are we sure we want to rename this file? To me its values look more like variables than settings, but if you all agree it makes more sense to call it settings then it's fine.

.azuredevops/platformPipelines/platform.updateReadMe.yml
```diff
@@ -13,7 +13,7 @@ trigger:
       - 'modules/**/deploy.json'
 
 variables:
-  - template: '../../global.variables.yml'
+  - template: '../../settings.yml'
```
eriqua, line 16: Is this used in this pipeline?

.github/workflows/ms.analysisservices.servers.yml
```diff
@@ -25,14 +25,15 @@ on:
...
-  variablesPath: 'global.variables.yml'
+  variablesPath: 'settings.yml'
   modulePath: 'modules/Microsoft.AnalysisServices/servers'
   workflowPath: '.github/workflows/ms.analysisservices.servers.yml'
   AZURE_CREDENTIALS: ${{ secrets.AZURE_CREDENTIALS }}
   ARM_SUBSCRIPTION_ID: '${{ secrets.ARM_SUBSCRIPTION_ID }}'
   ARM_MGMTGROUP_ID: '${{ secrets.ARM_MGMTGROUP_ID }}'
   ARM_TENANT_ID: '${{ secrets.ARM_TENANT_ID }}'
   DEPLOYMENT_SP_ID: '${{ secrets.DEPLOYMENT_SP_ID }}'
+  TOKEN_NAMEPREFIX: '${{ secrets.TOKEN_NAMEPREFIX }}'
```
eriqua, line 36: Was it tested if the pipeline succeeds when the secret is not configured?

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
```diff
@@ -90,26 +96,28 @@ To lower the barrier to entry and allow users to easily define their own naming
...
+
     > **Note:** The value should be a 3-5 character long string like `cntso`. Longer strings are not recommended as they may conflict with Azure resource name length restrictions.
 
+    > **Note:** The CI pipelines automatically removes `localToken_` section from the name when processing the tokens replacement.
+
     > **Note:** We highly encourage you to use the 'Check namePrefix availability' script ([see the documentation here](./Getting%20started%20-%20Check%20NamePrefix%20availability)) to check if the intended resource name will be available, based on the provided prefix.
 
  For further information on the token replacement logic, please refer to the corresponding [Token replacement](./The%20CI%20environment%20-%20Token%20replacement) section.
 
+ If you do not prefer to set the `namePrefix` locally for your CI environment. You can optionally set it as a GitHub Secret or add it to the ADO variable group as the key `TOKEN_NAMEPREFIX` with value `your name prefix value` and that will take precedence over the one specified in the `settings.yml`.
```
eriqua, line 119: I'd be more opinionated in suggesting to use the github secret over the local one. This is especially for the contribution scenario since contributors won't need to change the nameprefix value as a last thing anymore (finally ❤️ )

settings.yml
MrMCake: I think we don't / cannot replace token in our pipeline yml file as it is loaded before we could process it (at least in Azure DevOps).
Fortunately, for the ACR it's not so important as it's a deployment that sticks (and hence has low to no risks of conflicts).

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
MrMCake: For contributors yes, but for anybody onboarding it in a customer scenario, I assume they'd go with the settings file as it is more transparent and descriptive.

MrMCake: While I appreciate the possibility to set the prefix in the secrets, to me it feels more like a change for us, rather than anybody else. The namePrefix is after all technically not a secret ^^

.azuredevops/modulePipelines/ms.aad.domainservices.yml
MrMCake: I think one could make a case either way as there are both settings (e.g., telemetry) as well as classic variables (e.g., default RG) in the file.. And I'm fine either way.
Maybe we can get some more opinions on that 😊

.azuredevops/platformPipelines/platform.updateReadMe.yml
ahmadabdalla: I haven't recognized any variables in the pipeline that uses variables from the existing 'global.variables.yml'. Don't mind taking it off, it is an outcome from a find and replace, and I would prefer not to modify existing changes outside of the scope of this PR, but thumbs up if that's the preference.

.azuredevops/platformPipelines/platform.wiki-sync.yml
```diff
@@ -16,7 +16,7 @@ trigger:
 
 variables:
   - group: 'PLATFORM_VARIABLES'
-  - template: '../../global.variables.yml'
+  - template: '../../settings.yml'
```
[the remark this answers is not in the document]
ahmadabdalla: I haven't recognized any variables in the pipeline that uses variables from the existing 'global.variables.yml'. Don't mind taking it off, it is an outcome from a find and replace, and I would prefer not to modify existing changes outside of the scope of this PR, but thumbs up if that's the preference.

.github/workflows/ms.analysisservices.servers.yml
ahmadabdalla: yes

.azuredevops/platformPipelines/platform.dependencies.yml
```diff
@@ -134,23 +134,23 @@ stages:
...
-      - template: /.azuredevops/pipelineTemplates/jobs.validateModuleDeployment.yml
-        parameters:
-          deploymentBlocks:
-            - path: $(dependencyPath)/$(resourceType)/parameters/mg.parameters.json
-              templateFilePath: $(templateFilePath)
-              displayName: Policy assignment (mg)
-            - path: $(dependencyPath)/$(resourceType)/parameters/sub.parameters.json
-              templateFilePath: $(templateFilePath)
-              displayName: Policy assignment (sub)
+  #- stage: deploy_pa
```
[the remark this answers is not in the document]
ahmadabdalla: resolved

.azuredevops/platformPipelines/platform.dependencies.yml
ahmadabdalla: yes. But need to run another set of tests after some major PRs being merged. Will update status

.github/workflows/platform.dependencies.yml
ahmadabdalla: yes. But need to run another set of tests after some major PRs being merged. Will update status

ahmadabdalla:
@eriqua @MrMCake I've placed a 'blocked' tag, until we find a way to solve for the dependency for image that is impacted by the `namePrefix` token being a secret and not in a local file

ahmadabdalla:
@segraef need your support to confirm local testing functionality works on macOS as we're introducing a new mandated PowerShell module `powershell-yaml`

segraef:
> @segraef need your support to confirm local testing functionality works on macOS as we're introducing a new mandated PowerShell module `powershell-yaml`

Looks good so far, except 

image

.azuredevops/modulePipelines/ms.aad.domainservices.yml
ahmadabdalla: @MariusStorhaug initially provided the idea here https://github.com/Azure/ResourceModules/issues/1363#issuecomment-1166508084 and I think it's a good name to have hence suggested it. It is kind of that centralized place for config which is typically referred to as settings :P

.azuredevops/platformPipelines/platform.updateReadMe.yml
ahmadabdalla: I would suggest removing the references in another clean up PR. @MrMCake thoughts?

.azuredevops/platformPipelines/platform.wiki-sync.yml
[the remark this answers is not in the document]
ahmadabdalla: I would suggest removing the references in another clean up PR. @MrMCake thoughts?

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
ahmadabdalla: Although tested multiple times across ADO and GHA, I have left the current local name prefix to be carml. Do you reckon we should convert it to something like `cntso` ? As we want to obscure our token to protect our CI env name?

ahmadabdalla:
Latest tests for Dependency pipeline:
ADO: https://dev.azure.com/carml/CARML/_build/results?buildId=3485&view=results
GHA: https://github.com/Azure/ResourceModules/actions/runs/2837916196

NOTE: Faillure is not related to this change. Known issue #1727

.github/workflows/ms.synapse.workspaces.yml
```diff
@@ -32,6 +32,7 @@ env:
   ARM_SUBSCRIPTION_ID: '${{ secrets.ARM_SUBSCRIPTION_ID }}'
   ARM_MGMTGROUP_ID: '${{ secrets.ARM_MGMTGROUP_ID }}'
   ARM_TENANT_ID: '${{ secrets.ARM_TENANT_ID }}'
+  TOKEN_NAMEPREFIX: '${{ secrets.TOKEN_NAMEPREFIX }}'
```
MrMCake, line 35: Variable path above not updated

.github/workflows/platform.dependencies.yml
```diff
@@ -1167,13 +1308,17 @@ jobs:
             '11.azfw.parameters.json',
             '12.bastion.parameters.json',
             '13.bastion.parameters.json',
-            'parameters.json'
+            'parameters.json',
```
MrMCake, line 1311: @eriqua seems like I was not the only one with the extension after all 😉

MrMCake: @ahmadabdalla, in @eriqua's name I'd ask you to remove tailing commas 😄

settings.yml
```diff
@@ -3,6 +3,26 @@ variables:
...
+  # See: https://github.com/Azure/ResourceModules/wiki/The%20library%20-%20Module%20design#telemetry
+  enableDefaultTelemetry: true
+
+  ######################################
+  # Local tokens settings
+  ######################################
+
+  # the 'localToken_' prefix will be removed from the key name when the pipelines run.
+  # e.g. if you have a token in your parameter file as <<customKey>>, then the token defined in this file looks like "localToken_customKey": 'value'
+  localToken_namePrefix: 'carml' # A 3-5 character length string, included in the resources names, is overridden if provided as GitHub Secret / ADO Variable
```
MrMCake, line 16: I guess we'll leave this empty?

utilities/pipelines/staticValidation/module.tests.ps1
```diff
@@ -1015,37 +1011,41 @@ Describe 'Deployment template tests' -Tag Template {
...
-                [string] $testFilePath,
-                [string] $parameterFileName,
-                [hashtable] $tokenSettings,
-                [string] $tokenName,
-                [string] $tokenValue,
-                [string] $moduleFolderName
-            )
-            $ParameterFileTokenName = -join ($tokenSettings.tokenPrefix, $tokenName, $tokenSettings.tokenSuffix)
-            $ParameterFileContent = Get-Content -Path $testFilePath
+        foreach ($parameterFileTokenTestCase in $parameterFileTokenTestCases) {
```
MrMCake, line 1031: The for each seems redundant as we provide the test cases. Further, last time I checked Pester does not support this setup anymore

utilities/tools/Test-ModuleLocally.ps1
```diff
@@ -10,7 +9,7 @@ subscription Id, principal Id, tenant ID and other parameters that need to be to
 .PARAMETER TemplateFilePath
 Mandatory. Path to the Bicep/ARM module that is being tested
 
-.PARAMETER ParameterFilePath
+.PARAMETER TestFilePath
```
MrMCake, line 12: I'd suggest to continue calling this parameter file path as that is what it is. We won't ever provide a template test file path through this parameter. In the rest of the CI we (should) only have used testfileapath if the underlying file could be a parameter file or a template file

utilities/tools/Test-ModuleLocally.ps1
```diff
@@ -139,7 +138,7 @@ function Test-ModuleLocally {
     begin {
         $repoRootPath = (Get-Item $PSScriptRoot).Parent.Parent
         $ModuleName = Split-Path (Split-Path $TemplateFilePath -Parent) -Leaf
-        Write-Verbose "Running local tests for [$ModuleName]"
+        Write-Verbose "Running Local Tests for $($ModuleName)"
```
MrMCake, line 141: Please undo. Not only are the extra brackets not required, but in my subjektiv opinion, the original brackets [ ]  make it easier to see which part of an output is the variable one

utilities/tools/Test-ModuleLocally.ps1
```diff
@@ -247,30 +251,31 @@ function Test-ModuleLocally {
...
             }
+
             try {
                 # Validate template
                 # -----------------
                 if ($ValidationTest) {
                     # Loop through test parameter files
                     foreach ($moduleTestFile in $moduleTestFiles) {
-                        Write-Verbose ('Validating module [{0}] with test file [{1}]' -f $ModuleName, (Split-Path $moduleTestFile -Leaf)) -Verbose
+                        Write-Verbose ('Validating Module [{0}] with test file [{1}]' -f $ModuleName, (Split-Path $moduleTestFile -Leaf)) -Verbose
```
MrMCake, line 261: Why do we change the casing? 😄 I see that also happened with a lot of variables that are not parameters. Search & Replace?

docs/wiki/The CI environment - Token replacement.md
```diff
@@ -39,37 +39,38 @@ These are tokens constructed from environment variables, which are defined in th
...
-      "value": "tokenValue",
-      "metadata":{
-        "description":"token description"
-      }
-    }
-  ]
-}
 ```
 
+> **Note:** The CI pipelines automatically removes `localToken_` section from the name when processing the tokens replacement. This means that your actual token name is `tokenA` and NOT `localToken_tokenA`.
```
MrMCake, line 50:
```suggestion
> **Note:** The CI pipelines automatically removes the `localToken_` prefix from the name when processing the tokens replacement. This means that your actual token name is `tokenA` and NOT `localToken_tokenA`.
```

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
```diff
@@ -92,26 +98,28 @@ To lower the barrier to entry and allow users to easily define their own naming
...
-        "name": "namePrefix",
-        "value": "<...>"
-    }
+    ```yml
+    localToken_namePrefix: 'cntso'
     ```
+
     > **Note:** The value should be a 3-5 character long string like `cntso`. Longer strings are not recommended as they may conflict with Azure resource name length restrictions.
 
+    > **Note:** The CI pipelines automatically removes `localToken_` section from the name when processing the tokens replacement.
```
MrMCake, line 115:
```suggestion
    > **Note:** The CI pipelines automatically removes the `localToken_` prefix from the name when processing the tokens replacement.
```

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
```diff
@@ -136,6 +144,7 @@ To use the environment's pipelines you should use the information you gathered d
 | `ARM_TENANT_ID` | `9734cec9-4384-445b-bbb6-767e7be6e5ec` | The tenant ID of the Azure Active Directory tenant to test-deploy modules in. |
 | `AZURE_CREDENTIALS` | `{"clientId": "4ce8ce4c-cac0-48eb-b815-65e5763e2929", "clientSecret": "<placeholder>", "subscriptionId": "d0312b25-9160-4550-914f-8738d9b5caf5", "tenantId": "9734cec9-4384-445b-bbb6-767e7be6e5ec" }` | The login credentials of the deployment principal used to log into the target Azure environment to test in. The format is described [here](https://github.com/Azure/login#configure-deployment-credentials). |
 | `PLATFORM_REPO_UPDATE_PAT` | `<placeholder>` | A private access token (PAT) with enough permissions assigned to it to push into the main branch. This PAT is leveraged by pipelines that automatically generate ReadMe files to keep them up to date. |
+| `TOKEN_NAMEPREFIX` | `<cntso>` | Optional. If you specify the name prefix token here, this will take precedence over the `localToken_namePrefix` specified in the `settings.yml`.  |
```
MrMCake, line 147: Didn't we describe it the opposite way in the issue?  -  If specified in the file it has priority? Also seems more obvious to me as it is more transparent & explicit. Or I'm mixing things up,

ahmadabdalla: Yes that was the initial suggestion from @eriqua, and yes I have flipped it around based on the demo sessions we had > 2 months ago, but I don't remember there were other opinions as I had it articulated here https://github.com/Azure/ResourceModules/issues/1440#issuecomment-1151033577. 

However, I think it depends on what the path we want to provide to customers as a starting point:

1) Prioritize the **remote** (secret/variable) `namePrefix` token. And also set the local one to a dummy value.
   -  Here we do not change the existing mechanism and support both models. So even we rename our current token locally to be called `cntso`. Then that is not impacting our CI. Also guide the users to change this value to their preferred token name when testing locally.
   - Even if users fork carml and not change the namePrefix locally, they don't impact our CI. 

  2) Prioritize the **local** `namePrefix` token. And also set it to `''` as default on file. Making it explicit change for users:
     - Here we are defaulting the user experience to the remote (secret/variable) `namePrefix` token via the getting started guide.
     - Here we have to remove our local `namePrefix` token and set the value to `''`. Also guide the users to go to the `settings.yml` and change the `''` to `'TheirNamePrefixTokenValue'` to allow for local testing.

3) Completely remove the **local** `namePrefix` token and default the user experience to the **remote** (secret/variable) `namePrefix` token

   - Remove support for the local file option and replace it with placeholder tokens for the sake of examples. 
   - Here we are introducing a major change, and we need to guide the users for local testing mechanism. Example: inject an additional token called 'namePrefix' with the Test-ModuleLocally script. 

After giving it additional thought as per the above and with the current implementation of option (1) with this PR. Would this require us to have another session discussing this? FYI, changing to option (2) [original description], is not hard as well.

utilities/tools/Test-ModuleLocally.ps1
ahmadabdalla: Yes maybe. Updated

utilities/tools/Test-ModuleLocally.ps1
ahmadabdalla: Not sure why this came out as a change as I didn't change it
image
But I updated it regardless

utilities/tools/Test-ModuleLocally.ps1
ahmadabdalla: I thought aligning terminologies align how our CI and Local tests work. I suspect we'd want to evolve this script when we move to the .bicep file testing mechanism? Not very opinionated here, as alignment was the only driver.

utilities/pipelines/staticValidation/module.tests.ps1
ahmadabdalla: Updated based on feedback and tested as well

.github/workflows/ms.synapse.workspaces.yml
ahmadabdalla: updated

settings.yml
ahmadabdalla: see other comment that describes the different options. I was thinking of renaming it to `cntso` instead if we set this as a placeholder for customers.

utilities/tools/Test-ModuleLocally.ps1
MrMCake: Agree. I think we'll remove the `ParameterFilePath`/`TestFilePath` parameter eventually (& adjust the logic accordingly).
For the time being though, I suggest to keep it as is. The alignment of the terminlology is good and important - but only in those cases where we actually refer to the same concept. 

To re-itertate: 
- In the CI environment we're referring to the `TestFilePath` as a placeholder for both classic Parameter Files (that would be invoked in combination with a template file) & the new Bicep Test Files (that would be invoked on their own). 
- In here, this parameter is used exclusively for the classic parameter files and you cannot actually use it for the new Bicep Test Files. Instead you'd provide only the `TemplateFilePath` which would point to on of the new Bicep Test Files (instead of a module template).

It's also how it is implemented in the `New-TemplateDeployment` & `Test-TemplateDeployment` scripts (i.e., they both use the `ParameterFilePath` parameter as it only inteded for Parameter-JSON files anyways).

I hope that made sense :D

utilities/tools/Test-ModuleLocally.ps1
MrMCake: It may have been a resolved conflict as I believe this line was changed in main recently. Anyhow, thanks :)

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
MrMCake: I saw we also have an open comment from @eriqua in this PR somwhere towards the top. She's definitely vouching for Option 1, while I'm of the opinion that Option 2 is the more intuitive way.

If you want we can also have another chat later today. 

I somewhat have this scenario in my head that somebody has an error, tries to troubleshoot it, sees that there is the name prefix specified in the `settings.yml` and does it no see reflected in Azure. Why? Well because there is also a secret this person cannot see that renders the local configuration meaningless. Good luck for that poor contributor to figure this out :D 

Also I can't stress enough that this feature is something we implement primarily for us to avoid conflicts with other forks. I'd expect that new joiners will go ahead and use the `settings.yml` file rather than the secret to setup the prefix as it's not only more transparent, but also simply more obvious. I wouldn't want to push people to use the secrets mechanism for something that isn't a secret for no good reason. 

So, to summarize - my opinion is that we should
- Support both the secret & `settings.yml`
- Use the secret on our end and leave the value in the `settings.yml` file empty
- Let the `settings.yml` file take priority if both are specified

If you have some time today, I'm also up for a chat. I'll also try and ping some of the other contributors to get their thoughts.

.azuredevops/platformPipelines/platform.updateReadMe.yml
MrMCake: Fine by me

.azuredevops/platformPipelines/platform.wiki-sync.yml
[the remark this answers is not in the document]
MrMCake: Fine by me

.github/workflows/platform.dependencies.yml
ahmadabdalla: I think I may have missed a couple of comments. Will address this. I don't remember adding this at all 😂

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
eriqua: I actually vouch for option 2, as described in the related issue. In particular, **I definitely vote for keeping the local token empty**.
- Keeping `carml` as an example, even if we adopt the remote secret, will keep generating issues with having the global unique names taken, which is the primary reason for which we're implementing this change, and I'm afraid all this great work would be useless.
- Keeping an example token, such as `cntso`, will solve the issue on our end, but will cause the same conflict among users adopting the CI environment, because let's consider that unfortunately not everyone follows the documentation in detail and many will stay with what is already set. It is what it is.

So my suggestion is to move to option 2, support both local and remote, keep the local token empty, agree that having the local one taking precedence if both are set makes sense, for the reasons described by @MrMCake  above.

The only point on which I don't 100% agree is that we're implementing this change only for us. Unfortunately the remote token in GitHub is only possible as a secret and not as a variable, as ADO allows. And I agree this makes setting a secret for something which is not a secret confusing. On the other hand, I believe the implementation is helping the contribution flow, not only our maintenance. Using the remote secret, contributors will be able to leave the local token empty, so avoiding to make a last change before opening PRs from their fork, something reviewers need to double check, which may be error prone especially when PRs are touching a lot of files. That's why in my other comment I suggest to be more vocal in leading users (especially contributors) to use the remote token as an option. And that's also why I'd like to find a way to avoid changing the container registry unique name in the settings.yml as well, but not in this PR.

ahmadabdalla: Sounds good @eriqua and thanks for jumping in as I know you're away. I know @MrMCake is also trying to get a hold of another opinion. Nonetheless, I am also in favour for option 2 now as it's much better clarity on the rationale. Will adjust and test to ensure we don't miss on something and also update the documentation explicitly call this out as requirement for customers for the CI to function (if they don't opt in for a remote/secret token), and perform local testing as well.

Will give it some thought on the continent registry after this PR.

MrMCake: @eriqua, thanks for jumping in. Then I misunderstood the other statement. Agree with everything you stated above.

settings.yml
ahmadabdalla: @MrMCake  @eriqua  Updated and set to empty as per option (2)

utilities/tools/Test-ModuleLocally.ps1
ahmadabdalla: @MrMCake Reverted to previous naming.

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
ahmadabdalla: @eriqua @MrMCake updated to perform the following:
- Out of the box cloning experience for customers will have the `localToken_namePrefix` set to `''`. 
- Customers will use the getting started guide to add their name prefix token in the settings.yml.
- Customers may also choose to leave it empty as the default, but the pipeline will fail if they do not provide the remote version of the namePrefix token.
- To perform local testing. Customers need to either: 
  - add an additional token to the Test-ModuleLocally script to include a name prefix.
  - OR update the `localToken_namePrefix` to include the name prefix they want to use.
- If customers leave the `localToken_namePrefix` in the settings.yml with their own value. The pipeline will prioritize it even if the remote/secret version is enabled.

So TL;DR I implemented option (2) 

@MrMCake can you please create the 'TOKEN_NAMEPREFIX' with our value in the ADO variable group so our ADO pipelines do not fail

MrMCake: @ahmadabdalla done, thanks

.azuredevops/pipelineTemplates/jobs.dependenciesModuleDeployment.yml
```diff
@@ -149,8 +149,8 @@ jobs:
                 Write-Verbose ('Using local tokens [{0}]' -f ($tokenMap.Keys -join ', ')) -Verbose
                 $ConvertTokensInputs.Tokens += $tokenMap
 
-                # Swap 'namePrefix' token if provided as a Azure DevOps variable
-                if(-not [String]::IsNullOrEmpty("$env:TOKEN_NAMEPREFIX")){
+                # Swap 'namePrefix' token if empty and provided as a Azure DevOps variable
+                if([String]::IsNullOrEmpty($ConvertTokensInputs.Tokens['namePrefix'])){
                   Write-Verbose 'Using [namePrefix] token from Azure DevOps Variable Groups' -Verbose
                   $ConvertTokensInputs.Tokens['namePrefix'] = "$env:TOKEN_NAMEPREFIX"
```
MrMCake, line 155: Wouldn't that need to be `= '$(TOKEN_NAMEPREFIX)'`?

MrMCake, line 155: Wouldn't that need to be `= '$(TOKEN_NAMEPREFIX)'` as per the tenant ID reference above?

MrMCake, line 155: Wouldn't that need to be = '$(TOKEN_NAMEPREFIX)' as per the tenant ID reference above?

MrMCake: If so, I believe the same would need to be changed for the other ADO pipelines

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
```diff
@@ -112,13 +112,13 @@ To update the `namePrefix`, perform the following steps:
...
 
     > **Note:** We highly encourage you to use the 'Check namePrefix availability' script ([see the documentation here](./Getting%20started%20-%20Check%20NamePrefix%20availability)) to check if the intended resource name will be available, based on the provided prefix.
 
  For further information on the token replacement logic, please refer to the corresponding [Token replacement](./The%20CI%20environment%20-%20Token%20replacement) section.
 
- If you do not prefer to set the `namePrefix` locally for your CI environment. You can optionally set it as a GitHub Secret or add it to the ADO variable group as the key `TOKEN_NAMEPREFIX` with value `your name prefix value` and that will take precedence over the one specified in the `settings.yml`.
+ If you do not prefer to set the `namePrefix` locally for your CI environment. You can optionally set it as a GitHub Secret or add it to the ADO variable group as the key `TOKEN_NAMEPREFIX` with value `your name prefix value` and that will only take precedence if the `localToken_namePrefix` is left empty in the `settings.yml`.
```
MrMCake, line 121:
```suggestion
If you want to avoid that the `namePrefix` is displayed in plain text your CI environment's settings file, you can optionally set it as a GitHub Secret or add it to the ADO variable group as the key `TOKEN_NAMEPREFIX` with a custom value. This value is only applied if the `localToken_namePrefix` in the `settings.yml` is left empty.
```

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
```diff
@@ -144,7 +144,7 @@ To use the environment's pipelines you should use the information you gathered d
...
 | `AZURE_CREDENTIALS` | `{"clientId": "4ce8ce4c-cac0-48eb-b815-65e5763e2929", "clientSecret": "<placeholder>", "subscriptionId": "d0312b25-9160-4550-914f-8738d9b5caf5", "tenantId": "9734cec9-4384-445b-bbb6-767e7be6e5ec" }` | The login credentials of the deployment principal used to log into the target Azure environment to test in. The format is described [here](https://github.com/Azure/login#configure-deployment-credentials). |
 | `PLATFORM_REPO_UPDATE_PAT` | `<placeholder>` | A private access token (PAT) with enough permissions assigned to it to push into the main branch. This PAT is leveraged by pipelines that automatically generate ReadMe files to keep them up to date. |
-| `TOKEN_NAMEPREFIX` | `<cntso>` | Optional. If you specify the name prefix token here, this will take precedence over the `localToken_namePrefix` specified in the `settings.yml`.  |
+| `TOKEN_NAMEPREFIX` | `cntso` | Optional. If you specify the name prefix token here, this will only take precedence when the `localToken_namePrefix` specified in the `settings.yml` is left empty.  |
```
MrMCake, line 147:
```suggestion
| `TOKEN_NAMEPREFIX` | `cntso` | Optional. If you specify the name prefix token here, it is only applied if the `localToken_namePrefix` specified in the `settings.yml` is left empty.  |
```

docs/wiki/The CI environment - Token replacement.md
```diff
@@ -67,7 +67,7 @@ The solution comes with one predefined local token `namePrefix`. This token is l
 
 1. By updating the value of `localToken_namePrefix` in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml), which then becomes `namePrefix` when the pipelines run.
 
-1. Creating a GitHub Secret / ADO variable called `TOKEN_NAMEPREFIX`, which then becomes `namePrefix` when the pipelines run. If created as a secret or ADO variable, it will take precedence over the one defined in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml).
+1. Creating a GitHub Secret / ADO variable called `TOKEN_NAMEPREFIX`, which then becomes `namePrefix` when the pipelines run. If created as a secret or ADO variable, it will only take precedence when the `localToken_namePrefix` in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml) is left empty.
```
MrMCake, line 70:
```suggestion
1. Creating a GitHub Secret / ADO variable called `TOKEN_NAMEPREFIX`, which then becomes `namePrefix` when the pipelines run. If created as a secret or ADO variable, it is only applied if the `localToken_namePrefix` in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml) is left empty.
```

utilities/tools/Test-ModuleLocally.ps1
```diff
@@ -34,7 +34,7 @@ Optional. A hashtable parameter that contains custom tokens to be replaced in th
 
 $TestModuleLocallyInput = @{
     TemplateFilePath           = 'C:\Microsoft.Network\routeTables\deploy.bicep'
-    TestFilePath               = 'C:\Microsoft.Network\routeTables\.test\parameters.json'
+    ParameterFilePath               = 'C:\Microsoft.Network\routeTables\.test\parameters.json'
```
MrMCake, line 37:
```suggestion
    ParameterFilePath          = 'C:\Microsoft.Network\routeTables\.test\parameters.json'
```

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
```diff
@@ -266,6 +275,7 @@ Based on the information you gathered in the [Azure setup](#1-configure-your-azu
 | `ARM_MGMTGROUP_ID` | `de33a0e7-64d9-4a94-8fe9-b018cedf1e05` | The group ID of the management group to test-deploy modules in. |
 | `ARM_SUBSCRIPTION_ID` | `d0312b25-9160-4550-914f-8738d9b5caf5` | The ID of the subscription to test-deploy modules in. |
 | `ARM_TENANT_ID` | `9734cec9-4384-445b-bbb6-767e7be6e5ec` | The tenant ID of the Azure Active Directory tenant to test-deploy modules in. |
+| `TOKEN_NAMEPREFIX` | `<cntso>` | Optional. If you specify the name prefix token here, this will take precedence over the `localToken_namePrefix` specified in the `settings.yml`.  |
```
MrMCake, line 278:
```suggestion
| `TOKEN_NAMEPREFIX` | `<cntso>` | Optional. If you specify the name prefix token here, it is only applied if the `localToken_namePrefix` specified in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml) is left empty.  |
```

.azuredevops/pipelineTemplates/jobs.dependenciesModuleDeployment.yml
ahmadabdalla: fixed

docs/wiki/Getting started - Scenario 2 Onboard module library and CI environment.md
ahmadabdalla: updated to 

 If you do not prefer to set the `namePrefix` locally for your CI environment. You can optionally set it as a GitHub Secret or add it to the ADO variable group as the key `TOKEN_NAMEPREFIX` and its value `your name prefix value`, and this is only applied if the `localToken_namePrefix` in the [settings.yml](https://github.com/Azure/ResourceModules/blob/main/settings.yml) is left empty.

utilities/tools/Test-ModuleLocally.ps1
ahmadabdalla: @MrMCake shall I resolve?

MrMCake (approved):

Issue #1273, opened 2022-04-13
Microsoft.Compute/virtualMachines - Encryption settings for subs not having feature enabled


MariusStorhaug:
## Description

VM module has the `encryptionAtHost` setting as `True` by default. However, in subscriptions where this feature is not enabled, the VM deployment fails.

This issue should end up with a decision and actions taken so we are sure that the consumer knows what to do when this setting is set to True. It should also end with a decision if we should set to true by default, relying on a discussion regarding what "secure by default" means.

MariusStorhaug:
Could also be as easy as guiding the consumer to run the following:

azure-pipeline.yml
```yaml
- stage: Prep
  displayName: Prepare subscription
  jobs:
  - job: PrepSub
    displayName: Prepare subscription
    steps:
    - task: AzurePowerShell@5
      inputs:
        azureSubscription: $(serviceConnection)
        ScriptType: 'InlineScript'
        azurePowerShellVersion: 'LatestVersion'
        pwsh: true
        Inline: |
          . .\Scripts\EnableResourceProviders\Enable-ResourceProviders.ps1 -Verbose
```

The `Enable-ResourceProviders.ps1` script
```pwsh
[CmdletBinding()]
param ()

$ResourceProviders = Get-Content -Path $PSScriptRoot\ResourceProviders.json -Raw | ConvertFrom-Json

foreach ($ResourceProvider in $ResourceProviders) {
    Register-AzResourceProvider -ProviderNamespace $ResourceProvider.name
    foreach ($feature in $ResourceProvider.features) {
        Register-AzProviderFeature -ProviderNamespace $ResourceProvider.name -FeatureName $feature
    }
}
```

The `ResourceProviders.json` file:
```json
[
    {
        "name": "Microsoft.Compute",
        "features": [
            "EncryptionAtHost"
        ]
    }
]
```

MrMCake:
I know it's a first draft, but if we implement that (and I think it's cool) then we should be idempotent - i.e. only try to register if its not already enabled. Also we'd need to clarify what the impact on a pipeline's runtime is

MrMCake:
Related to #1048

eriqua:
Or pointing to the official 1 liner docs https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disks-enable-host-based-encryption-cli#prerequisites. 
Including the feature registration in the pipeline wouldn't anyway cover scenario 2 (module consumers not leveraging the CI environment).

[...]

PR #1537, opened 2022-06-16
[Modules] Container registry updated default value

prasanjeets:
# Description

> Applied all default values as per  NIST 800 

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation

# Checklist

- [ ] I did format my code

### DIFF
--- a/arm/Microsoft.ContainerRegistry/registries/deploy.bicep
+++ b/arm/Microsoft.ContainerRegistry/registries/deploy.bicep
@@ -12,7 +12,7 @@ param location string = resourceGroup().location
 @description('Optional. Array of role assignment objects that contain the \'roleDefinitionIdOrName\' and \'principalId\' to define RBAC role assignments on this resource. In the roleDefinitionIdOrName attribute, you can provide either the display name of the role definition, or its fully qualified ID in the following format: \'/providers/Microsoft.Authorization/roleDefinitions/c2f4ef07-c644-48eb-af81-4b1b4947fb11\'.')
 param roleAssignments array = []
 
-@description('Optional. Configuration Details for private endpoints.')
+@description('Optional. Configuration Details for private endpoints. Container registries should use private endpoints.')
 param privateEndpoints array = []
 
 @description('Optional. Tier of your Azure container registry.')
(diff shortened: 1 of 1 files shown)


arm/Microsoft.ContainerRegistry/registries/deploy.bicep
```diff
@@ -59,7 +59,7 @@ param retentionPolicyDays int = 15
   'enabled'
 ])
 @description('Optional. The value that indicates whether encryption is enabled or not.')
-param encryptionStatus string = 'disabled'
+param encryptionStatus string = 'enabled'
```
eriqua, line 62: Does enabling encryption require a specific feature to be enabled in the Azure subscription? In that case for consistency with other modules this need to be disabled by default.

MrMCake: We reeeeaally need this discussion again :D

prasanjeets: @eriqua : encryption is already part of Bicep code, we just enabling it. Moreover, also gone through with Bicep code doc for RSV, Doesn't seems to be any additional feature require at Azure subscription.

arm/Microsoft.ContainerRegistry/registries/deploy.bicep
```diff
@@ -12,7 +12,7 @@ param location string = resourceGroup().location
 @description('Optional. Array of role assignment objects that contain the \'roleDefinitionIdOrName\' and \'principalId\' to define RBAC role assignments on this resource. In the roleDefinitionIdOrName attribute, you can provide either the display name of the role definition, or its fully qualified ID in the following format: \'/providers/Microsoft.Authorization/roleDefinitions/c2f4ef07-c644-48eb-af81-4b1b4947fb11\'.')
 param roleAssignments array = []
 
-@description('Optional. Configuration Details for private endpoints.')
+@description('Optional. Configuration Details for private endpoints. Container registries should use private endpoints.')
```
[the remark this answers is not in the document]
prasanjeets: Done

arm/Microsoft.ContainerRegistry/registries/deploy.bicep
eriqua: @MrMCake I'm not sure I understand your comment. This is what the team decided *so far* (meaning it may change after further discussion) as per issue #1273 and applies to all resources in the same scenario. That doesn't mean we should disable encryption by default. It means we disable it in case something (like a feature) needs to be explicitly enabled on the subscription which is not enabled by default, causing the module deployment to fail otherwise. Hence my question. If that is not the case here, i.e. if no additional feature is required as per @prasanjeets comment, then I'm totally fine to have the value enabled by default.

eriqua (approved):
🆗

PR #1547, opened 2022-06-17
[Modules] Lock module and cross reference from RG

eriqua:
# Description

- Creating a new module for Authorization Locks (rg + subscription scope)
- Enabling testing the module at resource group scope. Currently it's not possible to continuously test on subscription scope, thus the code for the subscription part has been tested manually.
- Cross referencing locks from the Resource Group module
- Removing nested locks module from resource group
- Adding new RG resource to dependencies pipeline to which testing locks (we don't want the cannot delete lock on the validation-rg meanwhile potentially all module pipelines run in parallel)
   > Note Wiki update for additional dependency will be suggested directly in the Wiki review PR to avoid merge conflicts

| Pipeline |
| - |
| [Authorization: Locks](https://github.com/eriqua/ResourceModules/actions/workflows/ms.authorization.locks.yml) |
| [Resources: ResourceGroups](https://github.com/eriqua/ResourceModules/actions/workflows/ms.resources.resourcegroups.yml) |

# Type of Change

- [ ] Update to documentation

# Checklist

- [ ] I did format my code


arm/Microsoft.Authorization/locks/deploy.bicep
```diff
@@ -0,0 +1,55 @@
...
+    mode: 'Incremental'
+    template: {
+      '$schema': 'https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#'
+      contentVersion: '1.0.0.0'
+      resources: []
+    }
+  }
+}
+
+module lock_rg 'resourceGroup/deploy.bicep' = if (!empty(resourceGroupName) && !empty(subscriptionId)) {
```
MrMCake, line 40: I may be mistaken - but shouldn't there be 2 cases? 
- Lock for RG and
- Lock for Subscription

eriqua: Correct, I didn't add it since I couldn't find a way to test it. 
The cross-reference cannot be tested since we don't have a subscription module. 
For testing the subscription case within the lock module we could use our testing subscription, but I'd avoid it since we'd apply a canNotDelete lock on the same subscription in which potentially all the resources are tested (and removed) meanwhile.

MrMCake: Yes and no - we should be able to test it manually, which should be good enough - at least to have to code in the module there. The parameter file to run on a regular basis would then obviously not contain that test. A shame it is not possible to avoid the inhertiance :D

eriqua: Not sure I like to provide code which is not **continuously** tested. That's what I meant.

MrMCake: And we definitely should, once we have the capabilities available. Modules with features that cannot be continously tested in our environment (e.g. all those we don't have the dependencies for) should be avoided if we can - or in other words: we should try to do our best to actually test them. 

That said, if that's not possible we can only suggest to test manually - or fundamentally throw the correponding code out. So far we accepted code that we knew was working at a point in time instead of outright rejecting it because our CI environment is limited.

MrMCake (approved):

PR #1585, opened 2022-06-24
[Modules] DataFactory: Added private endpoints & customer-managed-key implementation

MrMCake:
# Description

- Added private endpoint parameter & cross-reference
- Added CMK implementation

| Pipeline |
| - |
| [DataFactory: Factories](https://github.com/Azure/ResourceModules/actions/workflows/ms.datafactory.factories.yml) |

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/arm/Microsoft.DataFactory/factories/.parameters/parameters.json
+++ b/arm/Microsoft.DataFactory/factories/.parameters/parameters.json
@@ -66,7 +66,7 @@
             "value": [
                 {
                     "subnetResourceId": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-001/subnets/<<namePrefix>>-az-subnet-x-005-privateEndpoints",
-                    "service": "amlworkspace"
+                    "service": "dataFactory"
                 }
             ]
         },
--- a/arm/Microsoft.DataFactory/factories/deploy.bicep
+++ b/arm/Microsoft.DataFactory/factories/deploy.bicep
@@ -175,8 +175,16 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' = {
   tags: tags
   identity: identity
   properties: {
-    repoConfiguration: bool(gitConfigureLater) ? null : json('{"type": "${gitRepoType}","accountName": "${gitAccountName}","repositoryName": "${gitRepositoryName}",${((gitRepoType == 'FactoryVSTSConfiguration') ? '"projectName": "${gitProjectName}",' : '')}"collaborationBranch": "${gitCollaborationBranch}","rootFolder": "${gitRootFolder}"}')
-    publicNetworkAccess: bool(publicNetworkAccess) ? 'Enabled' : 'Disabled'
+    repoConfiguration: bool(gitConfigureLater) ? null : union({
+        type: gitRepoType
+        accountName: gitAccountName
+        repositoryName: gitRepositoryName
+        collaborationBranch: gitCollaborationBranch
+        rootFolder: gitRootFolder
+      }, (gitRepoType == 'FactoryVSTSConfiguration' ? {
+        projectName: gitProjectName
+      } : {}), {})
+    publicNetworkAccess: publicNetworkAccess && empty(privateEndpoints) ? 'Enabled' : 'Disabled'
     encryption: !empty(cMKKeyName) ? {
       identity: {
         userAssignedIdentity: cMKUserAssignedIdentityResourceId
--- a/arm/Microsoft.DataFactory/factories/readme.md
+++ b/arm/Microsoft.DataFactory/factories/readme.md
@@ -58,7 +58,7 @@
 | `location` | string | `[resourceGroup().location]` |  | Location for all Resources. |
 | `lock` | string | `''` | `[, CanNotDelete, ReadOnly]` | Specify the type of lock. |
 | `managedVirtualNetworkName` | string | `''` |  | The name of the Managed Virtual Network. |
-| `privateEndpoints` | array | `[]` |  | Configuration Details for private endpoints. |
+| `privateEndpoints` | array | `[]` |  | Configuration Details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible. |
 | `publicNetworkAccess` | bool | `True` |  | Enable or disable public network access. |
 | `roleAssignments` | array | `[]` |  | Array of role assignment objects that contain the 'roleDefinitionIdOrName' and 'principalId' to define RBAC role assignments on this resource. In the roleDefinitionIdOrName attribute, you can provide either the display name of the role definition, or its fully qualified ID in the following format: '/providers/Microsoft.Authorization/roleDefinitions/c2f4ef07-c644-48eb-af81-4b1b4947fb11'. |
 | `systemAssignedIdentity` | bool | `False` |  | Enables system assigned managed identity on the resource. |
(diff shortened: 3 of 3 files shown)


modules/Microsoft.DataFactory/factories/deploy.bicep
```diff
@@ -155,8 +175,24 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' = {
...
+        accountName: gitAccountName
+        repositoryName: gitRepositoryName
+        collaborationBranch: gitCollaborationBranch
+        rootFolder: gitRootFolder
+      }, (gitRepoType == 'FactoryVSTSConfiguration' ? {
+        projectName: gitProjectName
+      } : {}), {})
+    publicNetworkAccess: publicNetworkAccess && empty(privateEndpoints) ? 'Enabled' : 'Disabled'
+    encryption: !empty(cMKKeyName) ? {
+      identity: {
```
eriqua, line 189: Other resources have 
```
identity: !empty(cMKUserAssignedIdentityResourceId) ? {
        userAssignedIdentity: cMKUserAssignedIdentityResourceId
      } : null
```

Would it make sense to align?

modules/Microsoft.DataFactory/factories/deploy.bicep
```diff
@@ -155,8 +175,24 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' = {
...
+    repoConfiguration: bool(gitConfigureLater) ? null : union({
+        type: gitRepoType
+        accountName: gitAccountName
+        repositoryName: gitRepositoryName
+        collaborationBranch: gitCollaborationBranch
+        rootFolder: gitRootFolder
+      }, (gitRepoType == 'FactoryVSTSConfiguration' ? {
+        projectName: gitProjectName
+      } : {}), {})
+    publicNetworkAccess: publicNetworkAccess && empty(privateEndpoints) ? 'Enabled' : 'Disabled'
```
eriqua, line 187: If publicNetworkAccess are explicitly set to true, should we disable it anyway if privateendpoints param is not empty?

eriqua: Also, can we avoid to set publicNetworkAccess to true by default? 
I'd suggest to update its declaration above like
```
@description('Optional. Whether or not public endpoint access is allowed for this resource.')
@allowed([
  ''
  'Enabled'
  'Disabled'
])
param publicNetworkAccess string = ''
```
and then set it here with something like
```suggestion
    publicNetworkAccess: !empty(publicNetworkAccess) ? any(publicNetworkAccess) : (!empty(privateEndpoints) ? 'Disabled' : null)
```
as done for other resources

modules/Microsoft.DataFactory/factories/.test/parameters.json
```diff
@@ -61,6 +61,23 @@
...
+        },
+        "privateEndpoints": {
+            "value": [
+                {
+                    "subnetResourceId": "/subscriptions/<<subscriptionId>>/resourceGroups/validation-rg/providers/Microsoft.Network/virtualNetworks/adp-<<namePrefix>>-az-vnet-x-001/subnets/<<namePrefix>>-az-subnet-x-005-privateEndpoints",
+                    "service": "dataFactory"
+                }
+            ]
+        },
+        "cMKUserAssignedIdentityResourceId": {
```
eriqua, line 73: Same as other PRs comment. Should we test CMK through a dedicated test `encr.parameters.json`?

modules/Microsoft.DataFactory/factories/deploy.bicep
```diff
@@ -65,6 +65,21 @@ param systemAssignedIdentity bool = false
...
+@description('Optional. The resource ID of a key vault to reference a customer managed key for encryption from.')
+param cMKKeyVaultResourceId string = ''
+
+@description('Optional. The name of the customer managed key to use for encryption.')
+param cMKKeyName string = ''
+
+@description('Optional. User assigned identity to use when fetching the customer managed key. If not provided, a system-assigned identity can be used - but must be given access to the referenced key vault first.')
+param cMKUserAssignedIdentityResourceId string = ''
+
+@description('Optional. The version of the customer managed key to reference for encryption. If not provided, the latest key version is used.')
```
eriqua, line 80: Same as other PR comment I'd move this parameter before the `cMKUserAssignedIdentityResourceId` one

MrMCake: done

modules/Microsoft.DataFactory/factories/deploy.bicep
```diff
@@ -155,8 +175,24 @@ resource dataFactory 'Microsoft.DataFactory/factories@2018-06-01' = {
...
+      }, (gitRepoType == 'FactoryVSTSConfiguration' ? {
+        projectName: gitProjectName
+      } : {}), {})
+    publicNetworkAccess: publicNetworkAccess && empty(privateEndpoints) ? 'Enabled' : 'Disabled'
+    encryption: !empty(cMKKeyName) ? {
+      identity: {
+        userAssignedIdentity: cMKUserAssignedIdentityResourceId
+      }
+      keyName: cMKKeyName
+      keyVersion: !empty(cMKKeyVersion) ? cMKKeyVersion : null
```
eriqua, line 193: Is this enough? Other implementations have 
```suggestion
      keyVersion: !empty(cMKKeyVersion) ? cMKKeyVersion : last(split(cMKKeyVaultKey.properties.keyUriWithVersion, '/'))
```
which requires defining the existing `cMKKeyVaultKey` resource above.
If this is enough let's try to align other resources with this one instead.

modules/Microsoft.DataFactory/factories/deploy.bicep
MrMCake: > If publicNetworkAccess are explicitly set to true, should we disable it anyway if privateendpoints param is not empty?

Unclear. The user wouldn't actually see that. So the user might be under the impression he set's it to `true` even though we set it to false under the hood. Not clear what our approach should be.

MrMCake: > Also, can we avoid to set publicNetworkAccess to true by default? I'd suggest to update its declaration above like
> 
> ```
> @description('Optional. Whether or not public endpoint access is allowed for this resource.')
> @allowed([
>   ''
>   'Enabled'
>   'Disabled'
> ])
> param publicNetworkAccess string = ''
> ```
> 
> and then set it here with something like
> 
> as done for other resources

As per the policy recommendation it should actually be `false` by default. However, I remember we had this exact discussion about another resource and the result was that we did not set it to false.

modules/Microsoft.DataFactory/factories/deploy.bicep
MrMCake: This resource automatically assumes latest if no version is provided. Most resources don't and need the extra KeyVaultKey `existing` reference

modules/Microsoft.DataFactory/factories/deploy.bicep
eriqua: Not sure I fully get the reply as I look at the parameter description ;)

```
@description('Optional. User assigned identity to use when fetching the customer managed key. If not provided, a system-assigned identity can be used - but must be given access to the referenced key vault first.')
param cMKUserAssignedIdentityResourceId string = ''
```

MrMCake: The description is borderline wrong and a copy-paste mistake. Fixing

MrMCake: done

modules/Microsoft.DataFactory/factories/deploy.bicep
MrMCake: done

modules/Microsoft.DataFactory/factories/.test/parameters.json
```diff
@@ -24,7 +24,7 @@
             }
         },
         "publicNetworkAccess": {
-            "value": true
+            "value": "Disabled"
         },
```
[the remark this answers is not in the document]
MrMCake: done

PR #1619, opened 2022-06-27
[Misc] Add automatic assignment to Pull-Requests

MariusStorhaug:
# Description

Adding auto assign automation to PRs

## Pipeline references

N/A

Proof is in the PR.

image

# Type of Change

- [ ] Update to documentation

# Checklist

- [x] I did format my code


.github/workflows/autoassign.yml
```diff
@@ -0,0 +1,12 @@
+name: 'Auto Assign PR'
```
MrMCake, line 1:
```suggestion
name: '.Platform: Auto Assign PR'
```

MrMCake, line 1: Rename file to: `platform.autoassign.yml`.

However, maybe we want to be a bit more descriptive and could even name it `platform.auto-assign-PR`

.github/auto_assign.yml
```diff
@@ -0,0 +1,2 @@
+addAssignees: author
```
MrMCake, line 1: Does this need to be in a specific folder path? Otherwise I'd suggest to move it into a subfolder of sorts. Alternatively we could name it slightly differently to be more descriptive. For example as per the workflow: `platform.auto-assign-PR.configuration`. Just want to avoid ending up with an increasingly long list of files without any reference to why they are there or what they do.

A shame we need a file in the first place.

.github/workflows/autoassign.yml
MariusStorhaug: Fixed.

MariusStorhaug: Fixed.

.github/auto_assign.yml
MariusStorhaug: Fixed. Let me know what you think.

MrMCake (approved):

Issue #1320, opened 2022-04-19
[Feature request]: Automate management group scope deployment clean up


eriqua:
### Description

Deployment quota often exceeding at management group scope.

Management group scope deployments fail when deployment count exceeds 800.

To overcome this, deployments that are no longer needed should be deleted from the history Ref [Management group limits](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#management-group-limits). This is currently being done manually.

The failure frequency increased after enabling telemetry settings, doubling up the number of deployments per pipeline run.

Error log example: https://github.com/Azure/ResourceModules/runs/6043758601?check_suite_focus=true#step:4:442

MrMCake:
I guess we could debate whether that is a bug, or an enhancement. But it's in any case a pain ;)

MrMCake:
To not run into API throttling limits, I suggest to not use the default PS-cmdlet (Remove-AzResourceGroupDeployment), nor the default [REST-API call](https://docs.microsoft.com/en-us/rest/api/resources/deployments/delete-at-management-group-scope), but instead use the bulkDelete used by the AzurePortal (with maximum 100 items at a time):

Request URL: https://management.azure.com/batch?api-version=2020-06-01
Request Method: POST
Header Content Type: Content-Type: application/json

Payload:
```json
{
    "requests": [
        {
            "content": {
                "EntityIds": [
                    "/providers/Microsoft.Management/managementGroups/<managementGroupId>/providers/Microsoft.Resources/deployments/pid-47ed15a6-730a-4827-bcb4-0fd963ffbd82-l6ql6gmfsjros",
                    "/providers/Microsoft.Management/managementGroups/<managementGroupId>/providers/Microsoft.Resources/deployments/roleAssignments-20220419T1104096929Z",
                    "/providers/Microsoft.Management/managementGroups/<managementGroupId>/providers/Microsoft.Resources/deployments/roleDefinitions-20220419T1104041183Z",
                    "/providers/Microsoft.Management/managementGroups/<managementGroupId>/providers/Microsoft.Resources/deployments/pid-47ed15a6-730a-4827-bcb4-0fd963ffbd82-wpnvthqx5ddns",
                    "/providers/Microsoft.Management/managementGroups/<managementGroupId>/providers/Microsoft.Resources/deployments/wpnvthqx5ddns-RoleDefinition-MG-Module",
[...]

PR #1622, opened 2022-06-28
[Utilities] Added ManagementGroup-Deployment removal script

MrMCake:
# Description

- Created a utility to remove any non-running deployment from a given Management-Group scope
- Uses batch processing, which let's you remove the deployments within a few seconds.
- This utility can later be used in issue #1320 , but until we automate it, it's useful to have around for manual execution

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/utilities/tools/Clear-ManagementGroupDeployment.ps1
+++ b/utilities/tools/Clear-ManagementGroupDeployment.ps1
@@ -1,0 +1,96 @@
+﻿
+<#
+.SYNOPSIS
+Bulk delete all deployments on the given management group scope
+
+.DESCRIPTION
+Bulk delete all deployments on the given management group scope
+
+.PARAMETER ManagementGroupId
+Mandatory. The Resource ID of the Management Group to remove the deployments for.
+
+.PARAMETER DeploymentStatusToExclude
+Optional. The status to exlude from removals. By default, we exclude any deployment that is in state 'running'.
+
+.EXAMPLE
+Clear-ManagementGroupDeployment -ManagementGroupId 'MyManagementGroupId'
+
+Bulk remove all 'non-running' deployments from the Management Group with ID 'MyManagementGroupId'
+#>
+function Clear-ManagementGroupDeployment {
+
+
+    [CmdletBinding(SupportsShouldProcess)]
+    param (
+        [Parameter(Mandatory = $true)]
+        [string] $ManagementGroupId,
+
+        [Parameter(Mandatory = $false)]
+        [string] $DeploymentStatusToExclude = 'running'
+    )
+
+    # Load used functions
+    . (Join-Path $PSScriptRoot 'helper' 'Split-Array.ps1')
+
+    $getInputObject = @{
+        Method  = 'GET'
+        Uri     = "https://management.azure.com/providers/Microsoft.Management/managementGroups/$ManagementGroupId/providers/Microsoft.Resources/deployments/?api-version=2021-04-01"
+        Headers = @{
+            Authorization = 'Bearer {0}' -f (Get-AzAccessToken).Token
+        }
+    }
+    $response = Invoke-RestMethod @getInputObject
+
+    if (($response | Get-Member -MemberType 'NoteProperty').Name -notcontains 'value') {
+        throw ('Fetching deployments failed with error [{0}]' -f ($reponse | Out-String))
+    }
+
+    $relevantDeployments = $response.value | Where-Object { $_.properties.provisioningState -ne $DeploymentStatusToExclude }
+
+    if (-not $relevantDeployments) {
+        Write-Verbose 'No deployments found' -Verbose
+        return
+    }
+
+    $relevantDeploymentChunks = , (Split-Array -InputArray $relevantDeployments -SplitSize 100)
+
+    Write-Verbose ('Triggering the removal of [{0}] deployments of management group [{1}]' -f $relevantDeployments.Count, $ManagementGroupId)
+
+    $failedRemovals = 0
+    $successfulRemovals = 0
+    foreach ($deployments in $relevantDeploymentChunks) {
+
+        $requests = $deployments | ForEach-Object {
+            @{ httpMethod            = 'DELETE'
+                name                 = (New-Guid).Guid # $_.properties.correlationId # "73c46fa8-dffe-4b9e-9360-cd0dc3c7b4d8",
+                requestHeaderDetails = @{
+                    commandName = 'HubsExtension.Microsoft.Resources/deployments.BulkDelete.execute'
+                }
+                url                  = '/providers/Microsoft.Management/managementGroups/{0}/providers/Microsoft.Resources/deployments/{1}?api-version=2019-08-01' -f $ManagementGroupId, $_.name
+            }
+        }
+
+        $removeInputObject = @{
+            Method  = 'POST'
+            Uri     = 'https://management.azure.com/batch?api-version=2020-06-01'
+            Headers = @{
+                Authorization  = 'Bearer {0}' -f (Get-AzAccessToken).Token
+                'Content-Type' = 'application/json'
+            }
+            Body    = @{
+                requests = $requests
+            } | ConvertTo-Json -Depth 4
+        }
+        if ($PSCmdlet.ShouldProcess(('Removal of [{0}] deployments' -f $requests.Count), 'Request')) {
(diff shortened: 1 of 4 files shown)


utilities/tools/Clear-ManagementGroupDeployment.ps1
```diff
@@ -0,0 +1,96 @@
...
+function Clear-ManagementGroupDeployment {
+
+
+    [CmdletBinding(SupportsShouldProcess)]
+    param (
+        [Parameter(Mandatory = $true)]
+        [string] $ManagementGroupId,
+
+        [Parameter(Mandatory = $false)]
+        [string] $DeploymentStatusToExclude = 'running'
```
eriqua, line 29: Is there a use case for which we want to remove running deployments? If not, I'd avoid this option, removing only completed deployments, maybe updating the function and .ps file names to reflect this (e.g. Clear-ManagementGroupCompletedDeployment)

utilities/tools/Clear-ManagementGroupDeployment.ps1
```diff
@@ -0,0 +1,96 @@
...
+    )
```
eriqua, line 30: I'd suggest to add a boolean parameter to allow skipping the removal of failed deployments, useful for troubleshooting. E.g.
``` 
        [Parameter(Mandatory = $false)]
        [bool] $SkipFailedDeployments = $true
```

utilities/tools/Clear-ManagementGroupDeployment.ps1
```diff
@@ -0,0 +1,96 @@
...
+    . (Join-Path $PSScriptRoot 'helper' 'Split-Array.ps1')
+
+    $getInputObject = @{
+        Method  = 'GET'
+        Uri     = "https://management.azure.com/providers/Microsoft.Management/managementGroups/$ManagementGroupId/providers/Microsoft.Resources/deployments/?api-version=2021-04-01"
+        Headers = @{
+            Authorization = 'Bearer {0}' -f (Get-AzAccessToken).Token
+        }
+    }
+    $response = Invoke-RestMethod @getInputObject
```
eriqua, line 42: Why not using `Invoke-AzRestMethod`? Cleaner code avoiding Headers and consistent with what we use for removal.

utilities/tools/Clear-ManagementGroupDeployment.ps1
```diff
@@ -0,0 +1,96 @@
...
+        }
+        if ($PSCmdlet.ShouldProcess(('Removal of [{0}] deployments' -f $requests.Count), 'Request')) {
+            $response = Invoke-RestMethod @removeInputObject
+
+            $failedRemovals += ($response.responses | Where-Object { $_.httpStatusCode -notlike '20*' }  ).Count
+            $successfulRemovals += ($response.responses | Where-Object { $_.httpStatusCode -like '20*' }  ).Count
+        }
+    }
+
+    Write-Verbose 'Outcome' -Verbose
```
eriqua, line 92: Wondering if useful to add an option to list removed deployment names + deployment status, maybe disabled by default. Not necessarily in this PR.

utilities/tools/Clear-ManagementGroupDeployment.ps1
```diff
@@ -0,0 +1,96 @@
...
+
+    Write-Verbose ('Triggering the removal of [{0}] deployments of management group [{1}]' -f $relevantDeployments.Count, $ManagementGroupId)
+
+    $failedRemovals = 0
+    $successfulRemovals = 0
+    foreach ($deployments in $relevantDeploymentChunks) {
+
+        $requests = $deployments | ForEach-Object {
+            @{ httpMethod            = 'DELETE'
+                name                 = (New-Guid).Guid # $_.properties.correlationId # "73c46fa8-dffe-4b9e-9360-cd0dc3c7b4d8",
```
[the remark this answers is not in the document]
MrMCake: removed

utilities/tools/Clear-ManagementGroupDeployment.ps1
MrMCake: I tried - and failed. It returns error messages above subscriptions not being set - which is ironic as I want to execute actions on an above subscription-level. Troubleshooting didn't return and results so I went with the general and working `Invoke-RestMethod`.

utilities/tools/Clear-ManagementGroupDeployment.ps1
MrMCake: It is a good question. Though the more I think about it, the more I wonder what you could do with that kind of information. It's not like you could recover the deployments anyways after they're deleted ^^

eriqua: You sure won't but could find useful to know if past deployment you removed failed or not. Probably best would be to also return the deployment error in failure cases. BTW something we could discuss as an improvement of this already useful version.

utilities/tools/Clear-ManagementGroupDeployment.ps1
eriqua: disappointing but fine :)

utilities/tools/Clear-ManagementGroupDeployment.ps1
```diff
@@ -0,0 +1,96 @@
...
+.DESCRIPTION
+Bulk delete all deployments on the given management group scope
+
+.PARAMETER ManagementGroupId
+Mandatory. The Resource ID of the Management Group to remove the deployments for.
+
+.PARAMETER DeploymentStatusToExclude
+Optional. The status to exlude from removals. Can be multiple. By default, we exclude any deployment that is in state 'running'.
+
+.EXAMPLE
```
eriqua, line 15: I'd suggest to add an example for `DeploymentStatusToExclude `usage

utilities/tools/Clear-ManagementGroupDeployment.ps1
eriqua: Now that it is an array, sure. I'm anyway still in doubt if we shouldn't keep failed deployments by default. I'd go for it but it's also true we could eventually reach the same 800 limit with failed deployments only. What do you think?

utilities/tools/Clear-ManagementGroupDeployment.ps1
```diff
@@ -0,0 +1,96 @@
+﻿
```
eriqua, line 1: We should probably add a wiki page for this as done for other utilities.

utilities/tools/Clear-ManagementGroupDeployment.ps1
MrMCake: Folks will eventually run of ouf deployments and need to remove failed ones too. However, I'm fine with keeping them by default until this limit is actually reached. I guess the pipeline we want as part of CARML could have a flag that deletes them too if a user reaaaally wants to.

MrMCake: done

utilities/tools/Clear-ManagementGroupDeployment.ps1
MrMCake: Agreed

utilities/tools/Clear-ManagementGroupDeployment.ps1
MrMCake: done

utilities/tools/Clear-ManagementGroupDeployment.ps1
MrMCake: Was thinking about this - but I couldn't find a place where to put it. It was easy with the original Wiki structure as we had the 'utilities' parent folder, but now?
I eventually ditched the idea as I figured that we'll eventually invoke it via a scheduled pipeline in the other issue and it's not like we documented the other pipeline scripts ^^. Thoughts?

eriqua: I think it naturally belongs to the CI environment, deployment validation in particular. It makes sense to be used only after deploying modules at mg scope.

MrMCake: Alternatively - Known issues...

MrMCake: Added

eriqua (approved):
🆗

PR #1582, opened 2022-06-23
[Modules] Redis cache module

itpropro:
# Description

Added redis cache module with the following ResourceModule provided defaults:

- Telemetry
- Resource locks
- Diagnostics settings
- RBAC

# Type of Change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)

# Checklist

- [X] I'm sure there are no other open Pull Requests for the same update/change
- [X] My corresponding pipelines / checks run clean and green without any errors or warnings
- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation (readme)
- [X] I did format my code

PR #1632, opened 2022-07-01
[Modules] Updated `DeploymentExamples` title & moved Bicep in front of JSON

MrMCake:
# Description

- Updated `DeploymentExamples` title (now also showing the title of the file, instead of only a number)
- Moved Bicep in front of JSON
- Excluded new linter rule (which demands to define properties not with quotes where possible)

TODO:
- [x] Underneath deployment examples header add a disclaimer: What do the example titles refer to, parameters are ordered, where is the cut to optional, etc.
- [x] In the examples themselves (where is the split)

| Pipeline |
| - |
| [AnalysisServices: Servers](https://github.com/Azure/ResourceModules/actions/workflows/ms.analysisservices.servers.yml) |

# Type of Change

Please delete options that are not relevant.

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Update to documentation

Issue #1564, opened 2022-06-21
[Feature Request]: Align parameter files naming and scenarios


eriqua:
### Description

Align on the different use cases we want to cover in our module tests and reflect that in the parameter naming convention. For instance:

- modules covering multiple services should cover all test scenarios
   - e.g. cognitive services cover both speech and face services. parameter files should cover both use cases, hence include `speech.parameters.json` and `face.parameters.json`
- always include a `min.parameters.json` to cover only required parameters. For modules covering multiple services this could be duplicated if different services lead to different implementations.
   - e.g. cognitive services cover both speech and face services. parameter files should cover both use cases, hence include `speech.min.parameters.json` and `face.min.parameters.json`
- all modules providing cmk functionality should include a `encr.parameters.json` (or `cmk.parameters.json`). For modules covering multiple services this could be duplicated if different services lead to different implementations.
   > **Discussion needed**: we could go for either a "always use` encr.parameters.json`" for consistency and to clearly show shorter specific examples in the readme or "use it only if required due to incompatibility with other settings" to instead try to include all possible features in the main `parameters.json`. The latter approach would probably be preferable considering the new dependencies approach. In either case update the Wiki as needed.
- all modules providing networking functionalities ahould include a `net.parameters.json` (or similar). For modules covering multiple services this could be duplicated if different services lead to different implementations.
   > **Discussion needed**: same as above.

---

Update: this is now related to main.test.bicep wrapper modules

eriqua:
Related to #401 which would be the implementation out of this discussion. I'd also propose to rename current `max.parameters.json` (analysys services and API management) to `parameters.json` to have consistency throughout the repo for the main test file. This main test file is a good candidate to be used for the public Bicep registry contribution, so it makes sense to have a consistent naming for that one.

rahalan:
Wiki also needs to be updated

AlexanderSehr:
I wonder if this is still relevant given the AVM specs. Thoughts @eriqua?

eriqua:
@AlexanderSehr I think this is more related to the implementation than it is on the specs definition. I wish that was already the case in AVM. Starting from "the defaults test should only include required parametes" 😄 
We're still not covering the above, but specs are covering almost all already. However the gap is more on making sure that related AVM specs are fullfilled than about improving the specs coverage.

PR #1635, opened 2022-07-04
[Modules] Redis cache module

MrMCake:
# Description

Based on original PR by @itpropro #1582 

Added redis cache module with the following ResourceModule provided defaults:

- Telemetry
- Resource locks
- Diagnostics settings
- RBAC

## Pipeline reference

| Pipeline |
| - |
| [Build Status](https://dev.azure.com/carml/CARML/_build/latest?definitionId=103&branchName=users%2Fitpropro%2Fredis) |

# Type of Change

# Checklist

- [X] I did format my code


modules/Microsoft.Cache/redis/.bicep/nested_rbac.bicep
```diff
@@ -0,0 +1,61 @@
...
+])
+param principalType string = ''
+
+@sys.description('Optional. The description of the role assignment.')
+param description string = ''
+
+var builtInRoleNames = {
+  Owner: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '8e3af657-a8ff-443c-a75c-2fe8c4bcb635')
+  Contributor: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'b24988ac-6180-42a0-ab88-20f7382dd24c')
+  Reader: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'acdd72a7-3385-48ef-bd42-f606fba81ae7')
```
eriqua, line 27: Why were quotes removed for this 3 roles + the `masterreader` below?

modules/Microsoft.Cache/redis/deploy.bicep
```diff
@@ -0,0 +1,262 @@
...
+])
+@description('Optional. Specify the type of lock.')
+param minimumTlsVersion string = '1.2'
+
+@allowed([
+  'Disabled'
+  'Enabled'
+])
+@description('Optional. Control permission for data plane traffic coming from public networks while private endpoint is enabled.')
+param publicNetworkAccess string = 'Enabled'
```
eriqua, line 43: I've opened a new issue #1660 (based on issue #819) to align this to all modules as the implementation is still quite inconsistent. Suggested to align to what done for cognitive services:
```suggestion
@description('Optional. Whether or not public network access is allowed for this resource. For security reasons it should be disabled. If not specified, it will be disabled by default if private endpoints are set.')
@allowed([
  ''
  'Enabled'
  'Disabled'
])
param publicNetworkAccess string = ''
```

modules/Microsoft.Cache/redis/deploy.bicep
```diff
@@ -0,0 +1,262 @@
...
+
+resource redisCache 'Microsoft.Cache/redis@2021-06-01' = {
+  name: name
+  location: location
+  tags: tags
+  identity: identity
+  properties: {
+    enableNonSslPort: enableNonSslPort
+    minimumTlsVersion: minimumTlsVersion
+    publicNetworkAccess: publicNetworkAccess
```
eriqua, line 192: Also as per previous comment
```suggestion
    publicNetworkAccess: !empty(publicNetworkAccess) ? any(publicNetworkAccess) : (!empty(privateEndpoints) ? 'Disabled' : null)
```

modules/Microsoft.Cache/redis/deploy.bicep
```diff
@@ -0,0 +1,262 @@
...
+output resourceGroupName string = resourceGroup().name
+
+@description('Redis host name.')
+output hostName string = redisCache.properties.hostName
+
+@description('Redis SSL port.')
+output sslPort int = redisCache.properties.sslPort
+
+@description('The full resource ID of a subnet in a virtual network where the Redis cache was deployed in.')
+output subnetId string = !empty(subnetId) ? redisCache.properties.subnetId : ''
```
eriqua, line 259: Are all these additional outputs intended?

modules/Microsoft.Cache/redis/.test/full.parameters.json
```diff
@@ -0,0 +1,61 @@
+{
```
eriqua, line 1: Should this be called parameter.json?

modules/Microsoft.Cache/redis/deploy.bicep
```diff
@@ -0,0 +1,262 @@
...
+@description('Optional. The SKU family to use. (C = Basic/Standard, P = Premium).')
+param family string = 'P'
+
+@allowed([
+  'Basic'
+  'Premium'
+  'Standard'
+])
+@description('Optional. The type of Redis cache to deploy.')
+param skuName string = 'Premium'
```
eriqua, line 92: Should `Premium` be the default sku? Wondering about the additional costs wrt `Basic` or `Standard`

modules/Microsoft.Cache/redis/.bicep/nested_rbac.bicep
itpropro: Following bicep best practices and to avoid bicep compiler and linter warning:

```
Warning prefer-unquoted-property-names: Property names that are valid identifiers should be declared without quotation marks and accessed using dot notation. [https://aka.ms/bicep/linter/prefer-unquoted-property-names]
```

modules/Microsoft.Cache/redis/deploy.bicep
itpropro: Makes sense for me. The only thing is that this has to be tested granularly, as I experienced multiple issue with RPs, handling null/empty values differently.

modules/Microsoft.Cache/redis/deploy.bicep
itpropro: Yes they are intended. I had usecases for these, and added them. As per design of the CARML modules, we have the base outputs plus the resource specific ones that makes sense and that don't expose security relevant information like access keys. That's at least how I understood it, maybe @MrMCake can elaborate :)

modules/Microsoft.Cache/redis/.test/full.parameters.json
itpropro: It's very inconsistent through all modules. As long as the reference in the Azure DevOps pipeline definition is correct this should be fine.
Can't wait to switch to bicep based parameter files better sooner than later 😉

modules/Microsoft.Cache/redis/deploy.bicep
itpropro: @MrMCake explained that by the principle of secure defaults, the default SKU should always be one that doesn't limit security relevant features such as managed identities, public endpoints, etc..

modules/Microsoft.Cache/redis/.test/full.parameters.json
eriqua: Hey @itpropro, I'm with you on bicep test files 😄 However, consider a test file naming pattern will still be needed once moving to .bicep tests, since the plan is to still use different files for different use cases. 
You're right about inconsistency, that's why we opened an issue to discuss that (see issue #1564). For the moment, I'd suggest to rename the most complete parameter file as `parameters.json` which is the most commonly used throughout the library.

modules/Microsoft.Cache/redis/.bicep/nested_rbac.bicep
MrMCake: Please undo. This new recommendation will (for the time being) be disabled in PR #1632.

modules/Microsoft.Cache/redis/deploy.bicep
MrMCake: Hey @itpropro, I think you're referring to my comment on the Web PubSub module where even the security policy created an Audit if one uses the Free-Tier. For that one the reason was (I believe) that one can only deploy Private Endpoints using the Standard Tier.
Now regarding the Redis Cache module we have to again make a case by case evaluation. @itpropro , are you aware of any specific security features (like private endpoints) we can only enable on the premium plan? I will also try and check the policies...

MrMCake: Ok, so I wasn't able to find anything SKU related in any policy recommendations (e.g. [here](https://github1s.com/Azure/azure-policy/blob/master/built-in-policies/policyDefinitions/Cache/RedisCache_AuditSSLPort_Audit.json)). Further, I just clicked through the service creation and it seems you can create e.g. private endpoints with the Standard SKU without any troubles.

modules/Microsoft.Cache/redis/.test/full.parameters.json
MrMCake: @itpropro I agree, it's a bit of a mixed bag. Personally, I don't really care for the name of the file as each just represents a test case and as such has the sole pupose of evaluation certain aspects of the module. In case of, for example, CosmosDB we have 4 files for the 4 different data base types - with no `parameters.json` in sight. 
I can't however also argue with the fact that the most common name we have today is `parameters.json` if the parameter file is just a general one without the expectation to test all, or e.g. the minimal amount of parameters. (Sidenote: To add to the inconsistency, some modules also have a `max.parameters.json` :D )

modules/Microsoft.Cache/redis/.bicep/nested_rbac.bicep
MrMCake: done

modules/Microsoft.Cache/redis/deploy.bicep
```diff
@@ -0,0 +1,281 @@
...
+@description('Optional. Static IP address. Optionally, may be specified when deploying a Redis cache inside an existing Azure Virtual Network; auto assigned by default.')
+param staticIP string = ''
+
+@description('Optional. The full resource ID of a subnet in a virtual network to deploy the Redis cache in. Example format: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/Microsoft.{Network|ClassicNetwork}/VirtualNetworks/vnet1/subnets/subnet1.')
+param subnetId string = ''
+
+@description('Optional. A dictionary of tenant settings.')
+param tenantSettings object = {}
+
+@description('Optional. Configuration Details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible.')
```
eriqua, line 97:
```suggestion
@description('Optional. Configuration details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible.')
```

modules/Microsoft.Cache/redis/deploy.bicep
```diff
@@ -0,0 +1,281 @@
...
+
+@description('Optional. Specifies whether the non-ssl Redis server port (6379) is enabled.')
+param enableNonSslPort bool = false
+
+@allowed([
+  '1.0'
+  '1.1'
+  '1.2'
+])
+@description('Optional. Specify the type of lock.')
```
eriqua, line 35: Wrong description

MrMCake: Updated

PR #1679, opened 2022-07-16
[Fixes] Linter fixes

eriqua:
# Description

A few linter fixes

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation

# Checklist

- [ ] I did format my code


modules/Microsoft.Web/staticSites/readme.md
```diff
@@ -1,6 +1,6 @@
-# Static Web Sites `[Microsoft.Web/staticSites]`
+# Web Static Sites `[Microsoft.Web/staticSites]`
 
-This module deploys a Static Web Site.
+This module deploys a web static site.
```
MrMCake, line 3: Why the rename? Should the description be the meaningful manual part. The provider is indeed Web - but that is the case is it is the 'grouping'. However, when talking about a static site, I'd not refer to it as a 'web static site', but definitely 'Static Website', 'Static Web Site' or just 'Static Site', wouldn't you agree?

MrMCake: Also we should have this conversation about casing :D I don't know why we change it here as we have both upper & lower casing for Azure services all over the repository. If we decide for one thing or the other we have to go in and update all others.

eriqua: Honestly enough the rename is a trick for the linter check. However, I didn't find a specific reference to a static web site in the docs, that's why I thought the workaround would have been acceptable.

On the casing I agree, let's have the discussion first if that is useful. I'm just trying to align to what I see most commonly used throughout the repo when I face it.

eriqua: Agreed and updated

MrMCake (approved):

PR #1680, opened 2022-07-16
[Utilities] Updated Dependency fetch script & added references to the module docs 

MrMCake:
# Description

- Merged the 2 dependency scripts into a single one (& updated the docs)
- Introduced a new section to the `Set-ModuleReadMe` script that adds those dependencies (be it local or from a registry) into the module's readme

| Pipeline |
| - |

# Type of Change

- [ ] Update to documentation


docs/wiki/Getting started - Get module cross-references.md
```diff
@@ -1,4 +1,4 @@
-The `'Get-LinkedLocalModuleList'` function helps you with discovering cross-modules dependencies. It checks for any local module references in a given path. This can be useful to determine which modules' folder you'd need to keep in case you'd only want to cherry-pick certain modules from the library.
+The `'Get-LinkedModuleList'` function helps you with discovering cross-modules dependencies. It checks for any module references in a given path. This can be useful to determine which modules' folder you'd need to keep in case you'd only want to cherry-pick certain modules from the library.
```
MrMCake, line 1: We should probably use 'reference' instead of 'dependency'

modules/Microsoft.ApiManagement/service/readme.md
```diff
@@ -266,6 +267,14 @@ userAssignedIdentities: {
 
 - *None*
 
+## Cross-referenced modules
+
+This section gives you an overview of all local-referenced module files (i.e., other CARML modules that are referenced in this module) and all remote-referenced files (i.e., Bicep modules that are referenced from a Bicep Registry or Template Specs).
+
+| Dependency | Type |
```
MrMCake, line 274: Rename

modules/Microsoft.AAD/DomainServices/readme.md
```diff
@@ -8,6 +8,7 @@ This template deploys Azure Active Directory Domain Services (AADDS).
 - [Parameters](#Parameters)
 - [Considerations](#Considerations)
 - [Outputs](#Outputs)
+- [Cross-referenced modules](#Cross-referenced-modules)
```
ahmadabdalla, line 11: Debatable topic: Do you think we should find a better/easier word for this:

- Module also uses:
- Depends on:
- Module leverages:
- Disclaimer: This module relies on the following modules from the library:

docs/wiki/Getting started - Get module cross-references.md
```diff
@@ -13,15 +13,20 @@ The `'Get-LinkedLocalModuleList'` function helps you with discovering cross-modu
...
 # How it works
 
 When invoking the script:
 
-1. The function leverages the utility [Get-LinkedModuleList](#related-function-get-linkedmodulelist) to fetch all references implemented in the modules in a given path.
-1. The function filters these references down to only local references (i.e., cross-module references) and formats them to show a consistent '\<ProviderNameSpace\>/\<ResourceType\>' format.
-1. Finally, it prints the references to the invoking terminal, group by ResourceType.
+1. The function fetches all references implemented in the modules in a given path.
+1. The function formats them to show a consistent '\<ProviderNameSpace\>/\<ResourceType\>' format.
+1. Finally, depending on whether you provided the `PrintLocalReferencesOnly` switch statement, the function either
```
ahmadabdalla, line 24: Do we have to have this via a switch? Why not write it to the host regardless and the returned output is what we use for the one for the readme?

utilities/tools/Set-ModuleReadMe.ps1
```diff
@@ -303,6 +303,81 @@ function Set-OutputsSection {
...
+
+    if ($dependencies.Keys -contains 'localPathReferences' -and $dependencies['localPathReferences']) {
+        foreach ($reference in ($dependencies['localPathReferences'] | Sort-Object)) {
+            $SectionContent += ("| ``{0}`` | {1} |" -f $reference, 'Local reference')
+        }
+    }
+
+    if ($dependencies.Keys -contains 'remoteReferences' -and $dependencies['remoteReferences']) {
+        foreach ($reference in ($dependencies['remoteReferences'] | Sort-Object)) {
+            $SectionContent += ("| ``{0}`` | {1} |" -f $reference, 'Remote reference')
```
ahmadabdalla, line 364: Does remote reference here allude to a module that is not in this library but exists in a template spec / bicep registry? If so, is this something the user needs to have in the documentation? As it does not technically require them to do any action in terms of 'NOT' deleting the local referenced module vs. the remote one?

modules/Microsoft.ApiManagement/service/readme.md
```diff
@@ -266,6 +267,14 @@ userAssignedIdentities: {
 
 - *None*
 
+## Cross-referenced modules
+
+This section gives you an overview of all local-referenced module files (i.e., other CARML modules that are referenced in this module) and all remote-referenced files (i.e., Bicep modules that are referenced from a Bicep Registry or Template Specs).
+
+| Reference | Type |
+| :-- | :-- |
+| `Microsoft.ApiManagement/authorizationServers` | Local reference |
```
ahmadabdalla, line 276: There's another comment about this. Do we need to explicitily call out 'remote reference'? What impact does it have on the user as shouldn't the API version table already contain the entire information of other types?

MrMCake: As far as I'm concerened there are 2 good reasons:
- Information on external dependencies is always good. If you reference a public endpoint it's 'easy' but what if it is a non-punlic private registry/template spec? Would you not want to be aware as a consumer?
- What if you're in an environment without public internet access (not the most uncommon scenario)?

utilities/tools/Set-ModuleReadMe.ps1
MrMCake: To your first question, yes, exactly.
Regarding the second question, see comment below.

docs/wiki/Getting started - Get module cross-references.md
MrMCake: Don't really see the negative here. The good thing about this switch being disabled by default is the reduced vebosity that would otherwise pollute the pipeline log with likely unreadable content (if the wide).

modules/Microsoft.AAD/DomainServices/readme.md
MrMCake: 'Cross-referenced' was the term we used until now. I did add a disclaimer to the section itself (underneath the header).
Ain't easy to find a good name, that is meaningful and does not imply something wrong. For example, 'Depends On' isn't ideal, as this conflicts with our reference to resource dependencies.

modules/Microsoft.ApiManagement/service/readme.md
ahmadabdalla: Thanks for clarifying. I know we do not have those use cases in our CI, at least today. And if we are going to support this feature, we need good documentation to explain the different reference types with examples. I couldn't see it in our doco for cross reference.

docs/wiki/Getting started - Get module cross-references.md
ahmadabdalla: Wouldn't say its a negativity. Thought might be another way to not have an additional step the user has to do or think of. Maybe if we use format table to make it a prettier output, and I don't know how much it will pollute the pipeline given its usually 1 or 2 cross-referenced modules?

modules/Microsoft.ApiManagement/service/readme.md
MrMCake: We don't use them in our CI, that's correct - but we never not supported them. 
Consumers however may. This feature borderline only surfaces these references if you choose to leverage them. I don't think we have any plans to do it ourselves though.

docs/wiki/Getting started - Get module cross-references.md
MrMCake: The script actually returns all dependencies (ideally for all modules) and not just a single one. You can provide a path to a specific subfolder - but then it will also only find the dependencies within that subfolder.
I can also just remove the output if having it seems wrong. I figured it was useful - but would not provide us with any benefit in the pipeline. 
Note: This output (for all modules) would be produced for each Pester test run, as it runs the script as part of the Set-ModuleRedMe check.

modules/Microsoft.ApiManagement/service/readme.md
ahmadabdalla: As long as we have it as a feature. We should surface some wording about what does local reference vs remote reference mean and when would the consumer see them and why. Would be useful if you can add that please.

docs/wiki/Getting started - Get module cross-references.md
ahmadabdalla: To avoid having the output impacting the pester test. Would it be better to decouple the functionality in a way that both the user script and the pester script benefits from it? Seems sharing a functionality that is intended to be user facing but at the same time use it for a CI function. Thoughts?

MrMCake: I'm not sure I follow. I could however offer the following: I can remove the switch and instead just leave the output as a verbose message (removing the hardcoded verbose switch that is currently implemented). So if the user wants to see some output they could invoke the function with a `-Verbose` and if they want to mimic the exact behavior, invoke it like
`$null = Get-LinkedModuleList -Verbose`

modules/Microsoft.AAD/DomainServices/readme.md
eriqua: +1 for cross-referenced modules, maybe because I'm used to it. 
If we keep that I'd suggest to rename the utility from `Get-LinkedModuleList` to `Get-CrossReferencedModuleList`

modules/Microsoft.ApiManagement/service/readme.md
MrMCake: Isn't the dislaimer exactly that? 
> This section gives you an overview of all local-referenced module files (i.e., other CARML modules that are referenced in this module) and all remote-referenced files (i.e., Bicep modules that are referenced from a Bicep Registry or Template Specs).

modules/Microsoft.AAD/DomainServices/readme.md
MrMCake: Renamed

docs/wiki/Getting started - Get module cross-references.md
ahmadabdalla: Yes that works too. Will explain what I meant later. Thanks @MrMCake !

modules/Microsoft.ApiManagement/service/readme.md
ahmadabdalla: Ah ok sorry I didn't easily spot this as it's only visible when there's a cross reference. I actually meant in the cross reference wiki for someone reading that documentation.

docs/wiki/Getting started - Get module cross-references.md
MrMCake: done

modules/Microsoft.ApiManagement/service/readme.md
MrMCake: Adjusted the description in the docs slightly to align better with the readme

docs/wiki/Getting started - Get module cross-references.md
ahmadabdalla: thanks @MrMCake . I think you forgot to remove the switch from the wiki :D

MrMCake: Damn :D

ahmadabdalla (approved):
Thank you :) Approved

PR #1686, opened 2022-07-17
[CI Environment] Removed temporal no-restore flag following the CLI fix

MrMCake:
# Description

- Now that CLI version 2.38.0 was release to the MS-hosted agents, the bug that was introduced with 2.37.0 is resolved. As a result we can remove the temporal workaround we introduced in the past

| Pipeline |
| - |

# Type of Change

- [ ] Update to documentation


eriqua (approved):
🆗

PR #1706, opened 2022-07-21
[Wiki] Updated template-orchestrated solution examples

MrMCake:
# Description

- Updated template-orchestrated solution examples
  - Moved parameters into module deployments
  - Simplified references

# Type of Change

- [x] Update to documentation

### DIFF
--- a/docs/wiki/Solution creation.md
+++ b/docs/wiki/Solution creation.md
@@ -153,7 +153,7 @@ module nsg 'modules/Microsoft.Network/networkSecurityGroups/deploy.bicep' = {
   name: 'registry-nsg'
   scope: resourceGroup(resourceGroupName)
   params: {
-    name: 'LocalFilesDemoNsg'
+    name: 'defaultNsg'
   }
   dependsOn: [
     rg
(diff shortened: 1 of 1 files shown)


docs/wiki/Solution creation.md
```diff
@@ -267,28 +225,40 @@ module rg 'br/modules:microsoft.resources.resourcegroups:0.4.735' = {
...
+        networkSecurityGroupName: nsg.name
+      }
+      {
+        name: 'SecondarySubnet'
+        addressPrefix: '10.0.1.0/24'
+        networkSecurityGroupName: nsg.name
+      }
+    ]
   }
   dependsOn: [
```
eriqua, line 261: Isn't this indirectly achieved by subnets depending on nsg which in turns depends on rg?

MrMCake: Yes., but it would be transient and hence potentially harder to understand. Thoughts?

eriqua: I'd avoid code which is not necessary, but I don't find that hard to understand so maybe my opinion is biased. Besides that I don't have other thoughts

MrMCake: don't have a strong opinion about it - and you at least have an opinion, so I updated it now :D

eriqua (approved):
🆗

PR #1770, opened 2022-08-09
[Modules] Introduced several linter-ignore statements for false-positives

MrMCake:
# Description

- Introduced several linter-ignore statements for false-positives -> Parameters are flagged because they 'may' me a secret (and should have the @secure annotation), even if they are not. 
- Otherwise, these would be surfaced as warnings when building the template

# Type of Change

- [ ] Update to documentation


modules/Microsoft.AAD/DomainServices/deploy.bicep
```diff
@@ -57,13 +57,15 @@ param ntlmV1 string = 'Enabled'
   'Enabled'
   'Disabled'
 ])
+#disable-next-line secure-secrets-in-params
```
eriqua, line 60: Can we have a small paragraph in the module design wiki for linter and what we intend to ignore?

MrMCake: Can do. I guess we'll need to revisit it every other month as the PG keeps on adding rules

eriqua: Not sure if in this or a separated PR, probably we would need to structure this a little more, but I'd like to hear your thoughts. 

We customize Bicep linter settings on 2 levels:

- The bicepconfig.json file for which we have a dedicated wiki page here https://github.com/Azure/ResourceModules/wiki/The%20CI%20environment%20-%20Bicep%20configuration%20file
- The single line to disable false positives as introduced in this PR (and also similarly leveraged by other 2 PRs currently in review) as documented here https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter#silencing-false-positives, a page we should reference since the list of rules is constantly updated as you correctly pointed out. 

So in the Wiki we should either have them in one page or at least reference each other. I find it a bit confusing that the first is documented in the CI environment while the second is documented in the module design.

For the latter, as also suggested in the link above, we should `add a comment explaining why the rule does not apply to this line.` in each .bicep file.

MrMCake: They are in fact about 2 distinct use cases that are only connected by the fact the the `bicepconfig.json` file is able to configure a subset of linter rules.

That being said, I'll go ahead and merge the two regardless - and to do that rename the original docs page to not reference the 'bicep config file' at all, but rather a 'bicep config' in general.

MrMCake: done

eriqua: I'd still suggest to `add a comment explaining why the rule does not apply to this line` to each .bicep file where we are silencing false positives, as mentioned here https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter#silencing-false-positives

docs/wiki/The CI environment - Bicep configuration.md
```diff
@@ -0,0 +1,28 @@
...
+
+For a full list of available rules, please refer to the [official documentation](https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-config).
+
+> If you remove this file, Bicep uses default values
+
+The configuration applied in the CI environment can be found in the `bicepconfig.json` file in the root folder. Next to the corresponding settings you will also find the rational for its application.
+
+# Module-level Bicep configuration
+
+The Bicep DL is continuously improved and extended with additional capabilities. Of of them, the Bicep Linter, provides guidance around template design & best practices - and surfaces any findings as warnings. However, while it is a great feature, there can be cases where rules show false-positives - or are not addressed by us immediately. For these cases, we occasionally apply ignore tags such as `#disable-next-line secure-secrets-in-params` on a module level.
```
eriqua, line 26: What's DL?

MrMCake: My bad, should be DSL (Domain Specific Language). Updated it

modules/Microsoft.AAD/DomainServices/deploy.bicep
MrMCake: Done :)

eriqua (approved):
🆗

Issue #1583, opened 2022-06-23
[Feature Request]: Enable new dependency approach from a pipeline perspective


MrMCake:
### Description

This issue should track our progress to enable the new dependency approach documented als 'Variant 5' here (#329). 

This means
- The pipeline should (side-by-side with the original approach) support `.bicep` test files on top of the `.json` test files (both for GitHub & Azure DevOps). This is important as reworking the modules to the new dependency approach will take time and cannot happen as a big-bang
- The new `.bicep` test files will deploy not only the parent template (`module <resourceType> '../deploy.bicep'`), but also the dependencies required for the given test case
  > **Note:** We should not use our own modules to deploy dependencies but raw `resource` deployments. However, we may centralize common patterns (e.g., for dependency resources) in centralized files. 
- The `.parameter` folder should be renamed to something more appropriate. This should happen in the predecessor #1577.

![image](https://user-images.githubusercontent.com/5365358/175328386-161bd065-29e8-4abf-9b4b-dc915ceb9068.png)

Issue #1186, opened 2022-03-26
[Feature request]: Testing - Evaluate & implement PSRule


MrMCake:
As presented by @BernieWhite a few months back we should checkout PSRule for CARML

As a source we should check out the [blog](https://blog.tyang.org/2022/03/20/azure-pipeline-psrule-bicep-test) published by @tyconsulting

MrMCake:
I'd suggest to not use the GitHub/AzureDevOps task to allow users to test locally too. Similar to what we do for the Pester tests,

BernieWhite:
@MrMCake Is testing in Visual Studio Code tasks fine, or do we also need a `./tests.ps1` type file for people to execute from the root of the repo?

Setting `ps-rule.yaml` and `.vscode/tasks.json` should be fine. #161

eriqua:
See also discussion in issue #1580

eriqua:
Leveraging this issue as a starting point for a PoC to integrate PSRule in the CARML CI environment pipelines.
The goal is to enable it for at least 3 modules already converted to the new dependencies approach (see issue #1791)

1. Pre-flight - Before resources are deployed from templates.
	- CARML PSRule (sub)set: 
		- Identify if any baseline rule does not apply to CARML and a customized set is required. 
		- Scope: identify if the test target should be repo/module/test specific or a combination of the 3.
	- Pester + PSrule:
		- Integrate current static validation with PSRule, identify any overlap and adjust the set of tests/rules accordingly.
	- (Bonus - if time allows) Custom rules
		- Inspect the need for custom rules and develop accordingly.
	- Run and generate output (see ref)
	- Act on results accordingly (e.g. update module)

2. In-flight - After resources are deployed to an Azure subscription.
	- Verify secure by default values (test specific)
	- Generate Output (see ref)
	- Act on results accordingly (e.g. update tests and default values)

3. Cross platform
	- Start with GH workflows and extend to ADO pipelines
	- (Bonus - if time allows) ARM/JSON converted modules


Ref:

- 	Main docs: https://azure.github.io/PSRule.Rules.Azure/about/
- 	GH repo: https://github.com/Azure/PSRule.Rules.Azure
- 	List of rules by resource type: https://azure.github.io/PSRule.Rules.Azure/en/rules/resource/
- 	Generating output: https://microsoft.github.io/PSRule/v2/analysis-output/ and https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/

[...]

PR #1773, opened 2022-08-10
PsRule GitHub action workflow

ms-sambell:
# Description

This code is not meant to be merged back into the ResourceModules repository. This fork is for a demo being prepared for the GeekReady conference. 

This PR adds a PS Rule IAC scan for the Bicep modules and published the results to a Log Analytics workspace. This data can then be visualised with an Azure Workbook to track IAC ratings against the WAF pillars.

### DIFF
--- a/.github/workflows/psRule.yml
+++ b/.github/workflows/psRule.yml
@@ -1,0 +1,27 @@
+name: PsRule - Analyze templates
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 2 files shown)


MrMCake:
Hey @ms-sambell cool that you played around with it. We also have an issue #1186 for that that we plan to implement once the effort surrounding #1583 are implemented

PR #1844, opened 2022-09-01
[Modules] Updated virtualNetworkGateways to support AAD VPN

ChrisSidebotham:
# Description

Updated module to support AAD Configuration for P2S VPN. Pipeline currently failing on resource removal of teh original vpn.parameters.json test file. New test file added `aadvpn.parameters.json` for AAD Configuration VPN and running successfully.

| Pipeline |
| - |
| [Network: VirtualNetworkGateways](https://github.com/Azure/ResourceModules/actions/workflows/ms.network.virtualnetworkgateways.yml) |

# Type of Change

- [x] Update to documentation

# Checklist

- [x] I did format my code


modules/Microsoft.Network/virtualNetworkGateways/deploy.bicep
```diff
@@ -251,7 +254,7 @@ var ipConfiguration = isActiveActiveValid ? [
   }
 ]
 
-var vpnClientConfiguration = {
+var vpnClientConfiguration = !empty(clientRootCertData) ? {
```
eriqua, line 257: Not sure I fully get the logic as I notice a nested condition on the same parameter at row 263. 

Also, it looks `vpnClientConfiguration` is set differently based on if either `clientRootCertData `or `vpnClientAadConfiguration` are empty. If that is the case for readability I'd probably suggest:
- to mention that in the parameter description (I think the 2 parameters cannot be both set)
- to create 2 different variables based on which value is empty
- to set the module parameter to the corresponding variable accordingly

ChrisSidebotham: Descriptions added to parameters to keep current functionality

eriqua: I'd still suggest to at least remove the condition on `vpnClientRootCertificates` at row 262, since we're checking on the empty value of `clientRootCertData`, which is redundant since the same condition is checked at row 256

ChrisSidebotham: Removed the duplicate check in latest commit

eriqua (approved):
🆗

PR #1884, opened 2022-09-04
[Modules] Updated VirtualNetwork to new dependency approach

MrMCake:
# Description

- Updated VirtualNetwork to new dependency approach

| Pipeline |
| - |
| [Network: VirtualNetworks](https://github.com/Azure/ResourceModules/actions/workflows/ms.network.virtualnetworks.yml) |

# Type of Change

- [ ] Update to documentation


modules/Microsoft.Network/virtualNetworks/.test/common/dependencies.bicep
```diff
@@ -0,0 +1,35 @@
...
+
+resource networkSecurityGroup 'Microsoft.Network/networkSecurityGroups@2022-01-01' = {
+    name: networkSecurityGroupName
+    location: location
+}
+
+@description('The resource ID of the created Route Table.')
+output routeTableResourceId string = routeTable.id
+
+@description('The resource ID of the created Route Table.')
```
eriqua, line 31:
```suggestion
@description('The resource ID of the created Network Security Group.')
```

modules/Microsoft.Network/virtualNetworks/.test/common/deploy.test.bicep
```diff
@@ -0,0 +1,128 @@
...
+// ========== //
+@description('Optional. The name of the resource group to deploy for a testing purposes.')
+@maxLength(90)
+param resourceGroupName string = 'ms.network.virtualnetworks-${serviceShort}-rg'
+
+@description('Optional. The location to deploy resources to.')
+param location string = deployment().location
+
+@description('Optional. A short identifier for the kind of deployment. Should be kept short to not run into resource-name length-constraints.')
+param serviceShort string = 'nvncom'
```
eriqua, line 14:
```suggestion
param serviceShort string = 'vnetcom'
```

modules/Microsoft.Network/virtualNetworks/.test/common/deploy.test.bicep
```diff
@@ -0,0 +1,128 @@
+targetScope = 'subscription'
+
+// ========== //
+// Parameters //
+// ========== //
+@description('Optional. The name of the resource group to deploy for a testing purposes.')
```
eriqua, line 6: Something to update in the utility

```suggestion
@description('Optional. The name of the resource group to deploy for testing purposes.')
```

modules/Microsoft.Network/virtualNetworks/.test/common/deploy.test.bicep
eriqua, line 14: Just as a side note, I see this is probably autogenerated, but `nvn` is not that clear for a vnet. I wouldn't block PRs on naming at this point, but added this discussion to the bucket of issue #2038

modules/Microsoft.Network/virtualNetworks/.test/vnetPeering/deploy.test.bicep
```diff
@@ -0,0 +1,67 @@
+targetScope = 'subscription'
+
+// ========== //
+// Parameters //
+// ========== //
+@description('Optional. The name of the resource group to deploy for a testing purposes.')
```
eriqua, line 6:
```suggestion
@description('Optional. The name of the resource group to deploy for testing purposes.')
```

modules/Microsoft.Network/virtualNetworks/.test/min/deploy.test.bicep
```diff
@@ -0,0 +1,40 @@
+targetScope = 'subscription'
+
+// ========== //
+// Parameters //
+// ========== //
+@description('Optional. The name of the resource group to deploy for a testing purposes.')
```
eriqua, line 6:
```suggestion
@description('Optional. The name of the resource group to deploy for testing purposes.')
```

modules/Microsoft.Network/virtualNetworks/.test/common/deploy.test.bicep
MrMCake: It's not autogenerated. I always followed the following approach:
- Of the provider (e.g., `Microsoft.EventGrid`), remove the 'Microsoft.' and take all capital letters. Here `eg`
- Of the resource type (e.g., `systemTopics`), do the same, starting with the first letter. Here: `st` 
- Last but not least, add the folder name (or a shortened version if necessary). E.g. `com`, or `min`.

Combined, I'd end up with something like
- `egstcom`
- `egstmin`

The idea here was that I'd have an easy time to identify where a resource was coming from because it's deterministic + there'd be a very low chance of conflict as I haven't seen a single combination of provider namespaces + resource type that would ever overlap with the same letters. Could be of course, but not yet.

I agree, we should cover that in #2038, but I'd very much like to keep it this way as it will ideally avoid confusion down the line.

eriqua: I could see the pattern, that's why I thought it was even generated. Anyway since there's a resource naming issue open for some time I'd still think it would be fair to discuss its solution.

eriqua (approved):
🆗

PR #1995, opened 2022-09-09
[Modules] Updated ManagementGroup to new dependency approach

MrMCake:
# Description

- Updated ManagementGroup to new dependency approach

| Pipeline |
| - |
| [Management: ManagementGroups](https://github.com/Azure/ResourceModules/actions/workflows/ms.management.managementgroups.yml) |

# Type of Change

- [ ] Update to documentation


modules/Microsoft.Management/managementGroups/deploy.bicep
```diff
@@ -28,14 +28,19 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
   }
 }
 
+resource parentManagementGroup 'Microsoft.Management/managementGroups@2021-04-01' existing = {
+  name: parentId
```
eriqua, line 32: Since the default value of `parentId` is empty, I guess we should either change the default value to `last(split(managementGroup().id` or add a condition to this block?

MrMCake: done & retested

eriqua (approved):
🆗

PR #2110, opened 2022-09-22
[Hackaton] Added PSRule Output to job summary

elanzel:
# Description
Updated vnetwork workflow and psrule output script


eriqua (approved):
🆗

PR #2134, opened 2022-09-26
[Utilities] Enabled no-param blocks

MrMCake:
# Description

- Enabled no-param blocks in Pester test files

| Pipeline |
| - |
| [Insights: ActionGroups](https://github.com/Azure/ResourceModules/actions/workflows/ms.insights.actiongroups.yml) |

# Type of Change

- [ ] Update to documentation


utilities/tools/Set-ModuleReadMe.ps1
```diff
@@ -166,7 +166,7 @@ function Set-ParametersSection {
         # 2. Create header including optional columns
         $newSectionContent += @(
             ('**{0} parameters**' -f $category),
-            (''),
+            '',
```
eriqua, line 169: We should run this on all module readmes to double check if updates would be introduced and avoid static validations to fail. Was that done?

MrMCake: Yes. By @ahmadabdalla in a PR that was merged yesterday. The above is just removing some redundant columns

MrMCake: Added a fix that cause the 2 changed files to disappear

eriqua (approved):
🆗

PR #2240, opened 2022-10-20
[Modules] Updated & extended ML Workspaces API version

MrMCake:
# Description

- Updated API version(s)
- Added 2 additional parameters 
- Added 2 additinoal allowed values for property

| Pipeline |
| - |
| [MachineLearningServices: Workspaces](https://github.com/Azure/ResourceModules/actions/workflows/ms.machinelearningservices.workspaces.yml) |

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/modules/Microsoft.MachineLearningServices/workspaces/deploy.bicep
+++ b/modules/Microsoft.MachineLearningServices/workspaces/deploy.bicep
@@ -133,6 +133,12 @@ param imageBuildCompute string = ''
 @sys.description('Conditional. The user assigned identity resource ID that represents the workspace identity. Required if \'userAssignedIdentities\' is not empty and may not be used if \'systemAssignedIdentity\' is enabled.')
 param primaryUserAssignedIdentity string = ''
 
+@sys.description('Optional. The service managed resource settings.')
+param serviceManagedResourcesSettings object = {}
+
+@sys.description('Optional. The list of shared private link resources in this workspace.')
+param sharedPrivateLinkResources array = []
+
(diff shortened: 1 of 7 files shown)


modules/Microsoft.MachineLearningServices/workspaces/deploy.bicep
```diff
@@ -131,6 +133,12 @@ param imageBuildCompute string = ''
 @sys.description('Conditional. The user assigned identity resource ID that represents the workspace identity. Required if \'userAssignedIdentities\' is not empty and may not be used if \'systemAssignedIdentity\' is enabled.')
 param primaryUserAssignedIdentity string = ''
 
+@sys.description('Optional. The service managed resource settings.')
```
eriqua, line 136: Should these new parameters be added to the common tests?

MrMCake: In theory yes - but I haven't had time to look into how they work. Just saw they were missing. I don't think it's too bad as we're just passing the value through without any logic - but if you think it should be required anyways I'll try to find time to figure out how I can get the PrivateLinks & CosmosDB in here.

eriqua: Let's postpone that outside of this PR, but let's create an issue for that. I'd expect we have the same for many modules, but maybe easier to at least track the ones we notice.

eriqua: Created #2247

eriqua (approved):
🆗

PR #2251, opened 2022-10-25
Adjusted implementation for testing

MrMCake:
# Description

- Proposed update to get rid of one file by moving a dependency

| Pipeline |
| - |
| [VirtualMachineImages: ImageTemplates](https://github.com/Azure/ResourceModules/actions/workflows/ms.virtualmachineimages.imagetemplates.yml) |


eriqua (commented):
@MrMCake, @fabmas if it is enough to assign the role at RG level instead of at subscription level I agree with this change

PR #2280, opened 2022-11-04
[Fixes] Applied a fix to handle inline quotations (e.g. for commands)

AlexanderSehr:
# Description

- Applied a fix to handle inline quotations (e.g. for commands)
- Tested the generation on all components without any undesired changes

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation


eriqua (approved):
🆗

PR #2282, opened 2022-11-04
[Utilities] Updated the deployment name generation to be more detailed

AlexanderSehr:
# Description

- Adjusted the deployment name logic to regenerate it each time a retry takes place (added with a `t#` in the end)
- Also incorporated the module's name in case the template's file path allows it

Examples

- Analysis Servies
  image
- Key Vault
  ```
  VERBOSE: Deploying with deployment name [KeyVault-vaults-min-20221104T2011351293Z-t1]
  VERBOSE: Deploying with deployment name [KeyVault-vaults-min-20221104T2011411996Z-t2]
  VERBOSE: Deploying with deployment name [KeyVault-vaults-min-20221104T2011437597Z-t3]
  ```

| Pipeline |
| - |
| [KeyVault: Vaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.keyvault.vaults.yml) (unrelated failing pipeline) |

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/utilities/pipelines/resourceDeployment/New-TemplateDeployment.ps1
+++ b/utilities/pipelines/resourceDeployment/New-TemplateDeployment.ps1
@@ -210,7 +210,7 @@ function New-DeploymentWithParameterFile {
         do {
             # Generate a valid deployment name. Must match ^[-\w\._\(\)]+$
             do {
-                $deploymentName = "$deploymentNamePrefix-$(-join (Get-Date -Format 'yyyyMMddTHHMMssffffZ')[0..63])-#$retryCount"
+                $deploymentName = "$deploymentNamePrefix-$(-join (Get-Date -Format 'yyyyMMddTHHMMssffffZ')[0..63])-t$retryCount"
             } while ($deploymentName -notmatch '^[-\w\._\(\)]+$')
 
             Write-Verbose "Deploying with deployment name [$deploymentName]" -Verbose
--- a/utilities/pipelines/resourceDeployment/New-TemplateDeployment.ps1
+++ b/utilities/pipelines/resourceDeployment/New-TemplateDeployment.ps1
@@ -160,10 +160,16 @@ function New-DeploymentWithParameterFile {
         }
         if ($templateFilePath -match '.*(\\|\/)Microsoft.+') {
             # If we can assume we're operating in a module structure, we can further fetch the provider namespace & resource type
-            $providerNamespace = ((($templateFilePath -split 'Microsoft\.')[1] -replace '\\', '/') -split '/')[0]
-            $resourceType = ((($templateFilePath -split 'Microsoft\.')[1] -replace '\\', '/') -split '/')[1]
+            $shortPathElem = (($templateFilePath -split 'Microsoft\.')[1] -replace '\\', '/') -split '/' # e.g., AppConfiguration, configurationStores, .test, common, deploy.test.bicep
+            $providerNamespace = $shortPathElem[0] # e.g., AppConfiguration
+            $providerNamespaceShort = ($providerNamespace -creplace '[^A-Z]').ToLower() # e.g., ac
 
-            $deploymentNamePrefix = "$providerNamespace-$resourceType-$deploymentNamePrefix"
+            $resourceType = $shortPathElem[1] # e.g., configurationStores
+            $resourceTypeShort = ('{0}{1}' -f ($resourceType.ToLower())[0], ($resourceType -creplace '[^A-Z]')).ToLower() # e.g. cs
+
+            $testFolderShort = Split-Path (Split-Path $templateFilePath -Parent) -Leaf  # e.g., common
+
+            $deploymentNamePrefix = "$providerNamespaceShort-$resourceTypeShort-$testFolderShort" # e.g., ac-cs-common
         }
 
         $DeploymentInputs = @{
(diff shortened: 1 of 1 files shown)


utilities/pipelines/resourceDeployment/New-TemplateDeployment.ps1
```diff
@@ -208,6 +208,14 @@ function New-DeploymentWithParameterFile {
         [int]$retryCount = 1
 
         do {
+            # Generate a valid deployment name. Must match ^[-\w\._\(\)]+$
+            do {
+                $deploymentName = "$deploymentNamePrefix-$(-join (Get-Date -Format 'yyyyMMddTHHMMssffffZ')[0..63])-t$retryCount"
```
eriqua, line 213: I like the approach! Can we test it for the module with longest deploymentNamePrefix? Just want to rule out that we don't have a situation for which the timestamp is cut off. Plus a small suggestion due to the new suffix
```suggestion
                $deploymentName = "$deploymentNamePrefix-$(-join (Get-Date -Format 'yyyyMMddTHHMMssffffZ')[0..60])-t$retryCount"
```

AlexanderSehr: Applied. But how did you come about the '60'? I was actually suprised these long deployment names worked as I figured there must be a limit of 64 characters, but nope. Also this limitation above - Don't know why we have it, because the timestamp has anyways 'only' 20 characters.
I unfortunately wasn't able to find out what the 'official' limit is - could however test, by pushing one module to the extreme and provoke an error message.

AlexanderSehr: It's actually worse - as the deployment name prefix can have any length, right? At least now the full name is:
`Provider + Resource Type + Test Folder Name + Timestamp + Iteration`.

- `KeyVault-vaults-min-20221104T2011351293Z-t1` has a length of 43
- `AnalysisServices-servers-common-20221104T2011351293Z-t1` has a length of 55

Didn't have a longer hat. So, if the max length is 64, I'd propose to flip the name and order it by importance. Something like (BEWARE simplified example):
```Powershell
$folderName = Split-Path $templateFilePath -LeafBase
$deploymentName = 't{0}-{1}-{2}-{3}-{4}' -f $retryCount, (Get-Date -Format 'yyyyMMddTHHMMss'), $folderName, $resourceType, $providerNamespace
```

Which would lead to e.g. (long example)

- 't1-20221107T131159-common-imageTemplates-VirtualMachineImages' with a length of 61

and here we could apply the addition of the [0..64] and cut in the end if necessary as I take the provider Namespace as the length important when it is 'shortened'. 

Thoughts?

eriqua: And luckily, we only deploy top level resources 😆 
I see the point and I agree with the priority. Only a bit concerned about the usability when dealing with troubleshooting.
I'm also thinking about reusing something similar to the `serviceshort`. So e.g.
`vmiitcom-20221107T131159-t1`
or
`vmiit-common-20221107T131159-t1`

What's your take on that? If sounds good but you think it may take some time to implement, we can always merge a first solution and iterate on that opening an issue for later.

AlexanderSehr: 'Time to implement' is relative - won't take much and I rather spent it now than revisiting the topic again. 

I would however, suggest not to go for the 'serviceShort' itself (i.e., not read it out of the file) as I know for a fact our script is used e.g. also by the AKS Accelerator to deploy the solution etc. - which isn't really in scope for us - but I think it would anyways be good to only implement CARML-specifics if absolutly necessary.

AlexanderSehr: So instead of taking that out of the file, we could generate the service short by going up the stack of folder names & taking every first & upper case letter + first 3 letters of parent folder. 
- e.g., `asscom-20221104T2011351293Z-t1`

Or e.g. do that for the Provider Namespace, but take the full resource Type
- e.g., `asservercom-20221104T2011351293Z-t1`

eriqua: Definitely agree on not grabbing that from the deployment template, but to generate it from the filepath instead, that's what I meant, sorry if not clear. 

Both the 2 latest proposals right above work for me 👍

AlexanderSehr: Figured as much. Just wanted to make sure

AlexanderSehr: Just updated it. Now it is:
| Path | Deployment Name | Length |
| - | - | - |
| `AppConfiguration\configurationStores\.test\common\deploy.test.bicep` | `ac-cs-common-20221107T2011153824Z-t1` | 36 |
| `AnalysisServices\servers\.test\min\deploy.test.bicep` | `as-s-min-20221107T2011139636Z-t1` | 32 |

eriqua: Looks awesome. The only doubt I still have is if we should keep the cut on the whole resource name or if instead put it at a different level (e.g. prefix or prefix + timestamp). 
The only chance I see for the deployment name to exceed 64 length is for the test folder to be more than 35 chars long, very unlikely but still possible. It's very minor and perhaps I'm overthinking this, if you think so I'm fine with merging this solution in.

AlexanderSehr: I think you're overthinking it ;) Imagine the folder name is 45 characters - even then, we only cut 10 characters of the name which in case of `as-s-min-20221107T2011139636Z-t1` would be `as-s-min-20221107T2011` which even then would still be the full name + data + hh + mm.

eriqua: And we'd lose the -t# suffix, which is the reason why we're implementing this fix. Anyway, as said, I don't want to block the PR due to that, but wanted to at least mention it :)

AlexanderSehr: ok, moving the T to the front of the Timestamp

AlexanderSehr: done

PR #2305, opened 2022-11-09
[Pipelines] Added scheduled pipelines for Subcription & Management-Group Deployment removals

AlexanderSehr:
# Description

- Added scheduled pipelines for Subcription & Management-Group Deployment removals
  - Excludes any deployment that is running/failed unless they're older than 14 days (new setting in `settings.yml`)
- Updated docs
- Implemented for both Azure DevOps & GitHub
- Updated 'get pipeline inputs' composite actions to be 'generic' and updated all pipelines using it

| Pipeline |
| - |
| [AnalysisServices: Servers](https://github.com/AlexanderSehr/ResourceModules/actions/workflows/ms.analysisservices.servers.yml) |
| [Build Status](https://dev.azure.com/carml/CARML/_build/latest?definitionId=118&branchName=users%2Falsehr%2F1320_cleanup) |
| [.Platform: Remove Deployments](https://github.com/alsehr/ResourceModules/actions/workflows/platform.deployment.removal.yml) |

# Type of Change

- [ ] Update to documentation


settings.yml
```diff
@@ -29,6 +29,13 @@ variables:
 
   moduleTestFilePath: 'utilities/pipelines/staticValidation/module.tests.ps1'
 
+  ######################################
+  # Removal settings
+  ######################################
+
+  # Deployments are only moved if they are not running or failed - AND - if they are not older than the below threshold
+  daysToKeepDeploymentsFor: 14
```
eriqua, line 37: This may be misleading for 2 reasons:
1. It is not clear we're talking about a dedicated pipeline and not about the removal state of the module validation pipeline
2. The variable name suggests we're keeping all deployments unless older than 14 days. If I understood correctly, it would be 
```suggestion
  # A scheduled pipeline deletes subscription and (optionally) management group level deployments which are neither in a running nor in a failed state. I addition, failed deployments are also removed if they are older than the below threshold
  daysToKeepFailedDeploymentsFor: 14
```

AlexanderSehr: Yes and no. We decided to keep 'running' & 'failed' deployments. There could potentially be more states and the function would allow you to specify them.
I get your point with the ask for a more meaningful name, but would suggest to not include any specifics like that. Maybe we can turn it on its head and use something like `maxDeploymentRetentionInDays`? What I try to express is, that beyond this point deployments are always deleted. Not 'just' failed ones, but all.

eriqua: I agree, neither are correct. Your latest suggestion is going in the right direction though. What I'd like to avoid is the wrong message for which we're deleting all deployments as part of the module validation pipeline and only after 14 days, which is what I would think by looking at the current name. To clarify which pipeline the setting is for, what would you think about adding different sections, e.g.,  general settings (if shared by all), module validation pipeline settings, platform maintenance settings?

AlexanderSehr: I guess we somewhat have these 'sections' already indicated by prefixes like `Publish: xzy` for a section header. I guess the fundamental 2 sections are `Common` & `Azure DevOps only` and I guess we should keep that. So whatever we do - it would be subsections and at least for me it would be fine to continue using these prefixes in the 'correct' order. Currently for example `validation` followed by `removal` and eventuall `publish`. Creative ideas are welcome :D

AlexanderSehr: I now introduced a section 'Platform pipeline settings'. What do you think? Figured 'removal' is probably a bit misleading as it could be mixed up with the module removal.

PR #2294, opened 2022-11-07
[Utilities] `Set-ModuleReadMe` Improved function identification regex

AlexanderSehr:
# Description

- Improved function identification regex in `Set-ModuleReadMe` script
- Re-ran generation on all templates
- Extended agent setup outpit
- Added PowerShell installation as agents have 7.2.7 installed while our scripts require 7.3.0

| Pipeline |
| - |
| [AnalysisServices: Servers](https://github.com/Azure/ResourceModules/actions/workflows/ms.analysisservices.servers.yml) |

# Type of Change

Please delete options that are not relevant.

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Update to documentation

PR #2315, opened 2022-11-11
[Utiltities] Updated `Set-ModuleReadMe` script to work with latest PowerShell version

AlexanderSehr:
# Description

- Updated `Set-ModuleReadMe` script to work with latest PowerShell version
- In a new version, `ConvertFrom-Json` returns now an 'OrderedHashtable' instead of a default Hashtable which can cause issues
- Also, re-ran the script on all modules

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/utilities/tools/helper/ConvertTo-OrderedHashtable.ps1
+++ b/utilities/tools/helper/ConvertTo-OrderedHashtable.ps1
@@ -93,7 +93,7 @@ function ConvertTo-OrderedHashtable {
                 }
 
                 # Case: Primitive data types
-                $primitiveElements = $JSONObject[$currentLevelKey] | Where-Object { $_.GetType().Name -notin @('Object[]', 'Hashtable') } | ConvertTo-Json | ConvertFrom-Json -AsHashtable -NoEnumerate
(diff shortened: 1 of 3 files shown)


utilities/tools/helper/ConvertTo-OrderedHashtable.ps1
```diff
@@ -87,13 +87,13 @@ function ConvertTo-OrderedHashtable {
                 }
 
                 # Case: Array of objects
-                $hashTableElements = $JSONObject[$currentLevelKey] | Where-Object { $_.GetType().Name -eq 'Hashtable' }
+                $hashTableElements = $JSONObject[$currentLevelKey] | Where-Object { $_.GetType().Name -in @('Hashtable', 'OrderedHashtable') }
```
eriqua, line 90: Should we merge with #2294 ?

AlexanderSehr: It serves a different purpose, but sure. Let me change the target branch

PR #2316, opened 2022-11-11
[Modules] Update SQL MI to new dependencies approach

eriqua:
# Description

- Converting to new dependencies approach
- Fixed dependency between keys and encryption object

| Pipeline |
| - |
| [Sql: ManagedInstances](https://github.com/Azure/ResourceModules/actions/workflows/ms.sql.managedinstances.yml) |

# Type of Change

- [ ] Update to documentation

# Checklist

- [ ] I did format my code


modules/Microsoft.Sql/managedInstances/.test/common/dependencies.bicep
```diff
@@ -0,0 +1,350 @@
...
+@description('Required. The name of the Managed Identity to create.')
+param managedIdentityName string
+
+@description('Required. The name of the Key Vault to create.')
+param keyVaultName string
+
+@description('Optional. The location to deploy resources to.')
+param location string = resourceGroup().location
+
+var sqlMiVnetAddressPrefix = '10.0.0.0/16'
```
AlexanderSehr, line 19: why the variable(s) if they're only used in one single location?

modules/Microsoft.Sql/managedInstances/.test/min/dependencies.bicep
```diff
@@ -0,0 +1,289 @@
...
+@description('Required. The name of the Network Security Group to create.')
+param networkSecurityGroupName string
+
+@description('Required. The name of the Route Table to create.')
+param routeTableName string
+
+@description('Optional. The location to deploy resources to.')
+param location string = resourceGroup().location
+
+var sqlMiVnetAddressPrefix = '10.0.0.0/16'
```
AlexanderSehr, line 13: Same as the other comment. If those variables are used only in a single location, I'd much rather put them there like all the other values

eriqua: The subnet prefix is widely used. The vnet prefix is related to the subnet one. If for any reason one would need to change their value, they would likely both be updated, so I suggest to keep them close.

modules/Microsoft.Sql/managedInstances/.test/common/dependencies.bicep
eriqua: Same as other comment

modules/Microsoft.Sql/managedInstances/.test/min/dependencies.bicep
AlexanderSehr: Hm OK. Personally, I feel it is confusing to introduce values in other places if not absolutely necessary, but I won't block the PR for it.

AlexanderSehr (approved):

Issue #2242, opened 2022-10-20
[Bug Report]: `Test-NamePrefixAvailability` not compatible with new dependencies approach


AlexanderSehr:
### Describe the bug

The script [Test-NamePrefixAvailability](https://github.com/Azure/ResourceModules/blob/main/utilities/tools/Test-NamePrefixAvailability.ps1) relies on the original JSON-Parameter files to find all names that must be unique and hence validated.

As we moved away from the JSON-Parameter files in favor of Bicep-Template files, we should also update this logic or remove its reference from the wiki as it will produce misleading results.

### To reproduce

Run the `Test-NamePrefixAvailability` function with for example a used storage account name.

rahalan:
1.) add a note in utility docs
2.) fix it

PR #2330, opened 2022-11-14
[Utilities] Fix for test name prefix availability

fblix:
# Description

This PR includes updates to the `Test-NamePrefixAvailability` script. (as per #2242)
Changes were necessary since there are no more parameter.json files but bicep files used for parameter testing.
A pipeline run is not attached since the script is to be run locally (see "Getting started" section in the wiki).

Happy to discuss my approach :)

| Pipeline |
| - |
| N/A |

# Type of Change

# Checklist

- [x] I did format my code


utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,123 +63,136 @@ function Test-NamePrefixAvailability {
...
-                else {
-                    Write-Warning "=> Warning: The storage account name [$storageAccountName] is not available. Please try a different prefix."
-                    $prefixAvailable = $false
+                'Microsoft.ContainerRegistry/registries' {
+                    $filter = 'registryName:'
                 }
+                'Microsoft.KeyVault/vaults' {
+                    $filter = 'keyVaultName:'
+                }
+                Default { Write-Error 'I dont like you.' }
```
eriqua, line 84: ❤️

fblix: I've updated this one to a more boring but better understandable error message.

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -34,6 +38,9 @@ function Test-NamePrefixAvailability {
         [Parameter(Mandatory = $true)]
         [string] $namePrefix,
 
+        [Parameter(Mandatory = $true)]
+        [string] $serviceShort,
```
AlexanderSehr, line 42: This parameter will become a problem as you won't have all 400 serviceShorts at hand when running the script ;)

Ideally, in the logic below, you should simply fetch the serviceShort from the file that you're anyways processing and test against the full name

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,123 +63,139 @@ function Test-NamePrefixAvailability {
...
-            # -----------------------
-            $storageAccountFiles = $parameterFiles | Where-Object { $_ -match 'Microsoft.Storage/storageAccounts' }
-            $storageAccountNames = $storageAccountFiles | ForEach-Object { (ConvertFrom-Json (Get-Content $_ -Raw)).parameters.name.value } | Where-Object { $null -ne $_ }
+        $storageAccountNames = @()
+        $containerRegistryNames = @()
+        $keyVaultNames = @()
 
-            $keyVaultFiles = $parameterFiles | Where-Object { $_ -match 'Microsoft.KeyVault/vaults' }
-            $keyVaultNames = $keyVaultFiles | ForEach-Object { (ConvertFrom-Json (Get-Content $_ -Raw)).parameters.name.value } | Where-Object { $null -ne $_ }
+        $parameterFiles = (Get-ChildItem -Path $repoRoot -Recurse -Filter 'deploy.test.bicep').FullName | ForEach-Object { $_.Replace('\', '/') }
```
AlexanderSehr, line 71:
```suggestion
        $parameterFiles = (Get-ChildItem -Path $repoRoot -Recurse -Filter 'deploy.test.bicep').FullName | Where-Object {
          Test-Path (Join-Path (Split-Path $_ -Parent) 'dependencies.bicep') # Currently we only need to consider files that have ResourceGroup resources
        } | ForEach-Object { 
          $_.Replace('\', '/') 
        }
```

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,123 +63,139 @@ function Test-NamePrefixAvailability {
...
-            foreach ($storageAccountName in $storageAccountNames) {
-                $path = '/subscriptions/{0}/providers/Microsoft.Storage/checkNameAvailability?api-version=2021-04-01' -f $subscriptionId
-                $requestInputObject = @{
-                    Method  = 'POST'
-                    Path    = $path
-                    Payload = @{
-                        name = $storageAccountName
-                        type = 'Microsoft.Storage/storageAccounts'
-                    } | ConvertTo-Json
+        foreach ($relevantResourceType in $relevantResourceTypes) {
```
AlexanderSehr, line 73: The loops should be switched for an improved runtime. I.e.:
1. Outer Loop: Go through the parameter files
2. Inner Loop: Go through the Resource Types
Reason being that you then loop infinitely less through the big nummber of parameter files over and over again (and read their content)

fblix: not sure I get your point correctly, but I tried switching the inner with the outer loop.

However the performance difference should be neglectable, right? The API calls are what take up 99% of the runtime of the script.

AlexanderSehr: Doesn't mean we shouldn't code by example ;)

fblix: true :D

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -34,6 +39,9 @@ function Test-NamePrefixAvailability {
         [Parameter(Mandatory = $true)]
         [string] $namePrefix,
 
+        [Parameter(Mandatory = $false)]
+        [string] $overwrittenServiceShort,
```
AlexanderSehr, line 43: Just to double-check  - what's the use case here if the service shot is always different (per file). Is it just for testing?

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,127 +64,164 @@ function Test-NamePrefixAvailability {
...
-                $path = '/subscriptions/{0}/providers/Microsoft.Storage/checkNameAvailability?api-version=2021-04-01' -f $subscriptionId
-                $requestInputObject = @{
-                    Method  = 'POST'
-                    Path    = $path
-                    Payload = @{
-                        name = $storageAccountName
-                        type = 'Microsoft.Storage/storageAccounts'
-                    } | ConvertTo-Json
+                # determine serviceshort default value if no parameter has been supplied
+                if (!$overwrittenServiceShort) {
```
AlexanderSehr, line 105:
```suggestion
                if (-not $overwrittenServiceShort) {
```

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,127 +64,164 @@ function Test-NamePrefixAvailability {
...
+                        Write-Warning "The entered resource Type '$_' could not be matched."
+                        continue
+                    }
+                }
+            
+                $temp = $null
 
-            $subscriptionId = (Get-AzContext).Subscription.Id
+                # determine if entry is of one of the resourceTypes using the filter variable
+                $temp = Get-Content -Path $parameterFile | ForEach-Object {
```
AlexanderSehr, line 100: The content of the file should be loaded once in the loop and not for each search (i.e. resource name, serviceshort, etc.)

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,127 +64,164 @@ function Test-NamePrefixAvailability {
...
-                $requestInputObject = @{
-                    Method  = 'POST'
-                    Path    = $path
-                    Payload = @{
-                        name = $storageAccountName
-                        type = 'Microsoft.Storage/storageAccounts'
-                    } | ConvertTo-Json
+                # determine serviceshort default value if no parameter has been supplied
+                if (!$overwrittenServiceShort) {
+                    $serviceShort = Get-Content -Path $parameterFile | ForEach-Object {
```
AlexanderSehr, line 106: The whole block can be shortened via
```PowerShell
// With pre-loaded file content array
$serviceShort = $fileContent | ForEach-Object { 
if ($_ -Match "^param serviceShort string = '(\w+)'$") {
    $matches[1]
    break 
}
```

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,127 +64,164 @@ function Test-NamePrefixAvailability {
...
-                        type = 'Microsoft.KeyVault/vaults'
-                    } | ConvertTo-Json
-                }
-                $response = ((Invoke-AzRestMethod @requestInputObject).Content | ConvertFrom-Json)
-                if ($response.nameAvailable) { Write-Host "=> The key vault name [$keyVaultName] is currently available" -ForegroundColor 'Green' }
-                else {
-                    Write-Warning "=> Warning: The key vault name [$keyVaultName] is not available. Please try a different prefix."
-                    $prefixAvailable = $false
+                if ($temp) {
+                    $temp = $temp.Split($filter)
```
AlexanderSehr, line 118: Also this could be solved via regex & group matchings if you're up to the task ;) would probably save you 10+ lines

utilities/tools/Test-NamePrefixAvailability.ps1
fblix: Mainly for testing yes. Wanted to provide the ability if someone has a specific serviceshort in his mind to test this here too.

utilities/tools/Test-NamePrefixAvailability.ps1
JPEasier: @AlexanderSehr Updated with your suggestion

utilities/tools/Test-NamePrefixAvailability.ps1
JPEasier: @AlexanderSehr good point.
@fblix I made some changes in your branch.

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -32,16 +27,26 @@ Test-NamePrefixAvailability @inputObject
...
+    Tokens     = @{
+        Location          = 'westeurope'
+        ResourceGroupName = 'validation-rg'
+        SubscriptionId    = '00000000-0000-0000-0000-000000000000'
+        ManagementGroupId = '00000000-0000-0000-0000-000000000000'
+        RemoveDeployment  = $false
+        deploymentSpId    = '00000000-0000-0000-0000-000000000000'
+    }
+}
+Test-NamePrefixAvailability @inputObject
```
AlexanderSehr, line 41: Please remove this test. Also, I recommend to usually put it AFTER the function as you'll otherwise run the old code ;)

utilities/tools/Test-NamePrefixAvailability.ps1
```diff
@@ -56,123 +69,155 @@ function Test-NamePrefixAvailability {
...
+                    Default {
+                        Write-Warning "The entered resource Type '$_' could not be matched."
+                        continue
+                    }
+                }
 
-            $acrFiles = $parameterFiles | Where-Object { $_ -match 'Microsoft.ContainerRegistry/registries' }
-            $acrNames = $acrFiles | ForEach-Object { (ConvertFrom-Json (Get-Content $_ -Raw)).parameters.name.value } | Where-Object { $null -ne $_ }
+                $temp = $null
+                $fileContent = Get-Content -Path $parameterFile
```
AlexanderSehr, line 103: This should not be loaded for each resourceType again

utilities/tools/Test-NamePrefixAvailability.ps1
AlexanderSehr: Let's save that for later to get the PR in

AlexanderSehr (approved):

Issue #2038, opened 2022-09-16
[Feature Request]: Updated dependencies approach - Test formatting


eriqua:
### Description

This issue is about agreeing on the format to apply to module test templates (deploy.test.bicep) and implement the changes as a bulk edit once all related PRs will be merged.

In particular, section names and spacing e.g. use `// Dependencies //` instead of generic `// Deployments //` for the dependencies section.

The issue is blocked by #1791 or at least by the merge of all currently open PRs.

PR #2400, opened 2022-12-02
[Modules] Updated variable usage in test cases & added docs recommendation

AlexanderSehr:
# Description

- Updated variable usage in test cases 
- Added docs recommendation

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/modules/Microsoft.Sql/managedInstances/.test/common/dependencies.bicep
+++ b/modules/Microsoft.Sql/managedInstances/.test/common/dependencies.bicep
@@ -16,7 +16,6 @@ param keyVaultName string
 @description('Optional. The location to deploy resources to.')
 param location string = resourceGroup().location
 
-var sqlMiVnetAddressPrefix = '10.0.0.0/16'
 var sqlMiSubnetAddressPrefix = '10.0.0.0/24'
 var sqlMiSubnetAddressPrefixString = replace(replace(sqlMiSubnetAddressPrefix, '.', '-'), '/', '-')
 
@@ -263,7 +262,7 @@ resource virtualNetwork 'Microsoft.Network/virtualNetworks@2022-01-01' = {
   properties: {
     addressSpace: {
       addressPrefixes: [
-        sqlMiVnetAddressPrefix
+        '10.0.0.0/16'
       ]
     }
     subnets: [
--- a/docs/wiki/The library - Module design.md
+++ b/docs/wiki/The library - Module design.md
@@ -567,6 +567,7 @@ In addition, they follow these file-type-specific guidelines:
     - If the module deploys a resource group level resource, the template should further have a `resourceGroupName` parameter and subsequent resource deployment. As a reference for the default name you can use `ms.<providerNamespace>.<resourceType>-${serviceShort}-test-rg`.
     - Each file should also provide a `location` parameter that may default to the deployments default location
   - It is recommended to define all major resource names in the `deploy.test.bicep` file as it makes later maintenance easier. To implement this, make sure to pass all resource names to any referenced module.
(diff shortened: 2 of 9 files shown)


AlexanderSehr:
Linked to #2038

modules/Microsoft.Sql/managedInstances/.test/common/dependencies.bicep
```diff
@@ -16,7 +16,6 @@ param keyVaultName string
 @description('Optional. The location to deploy resources to.')
 param location string = resourceGroup().location
 
-var sqlMiVnetAddressPrefix = '10.0.0.0/16'
```
eriqua, line 19: As mentioned when opening this PR, although I agree with the general recommendation, I suggest to keep this variable giving the strict correlation between vnet and subnet address space

AlexanderSehr: But then what is the rule? :D 

> Don't create artificial variables, unless absolutly necessary - or because of high correlation with another variable?

eriqua: "Absolutely necessary" is difficult to define, I'd go with something even more difficult to define: "Don't create artificial variables, unless it makes sense" 😆

AlexanderSehr: Love it. #needsFurtherDiscussion ;)

modules/Microsoft.AAD/DomainServices/.test/common/dependencies.bicep
```diff
@@ -15,21 +15,22 @@ param certDeploymentScriptName string
 
 var certPWSecretName = 'pfxCertificatePassword'
 var certSecretName = 'pfxBase64Certificate'
+var addressPrefix = '10.0.0.0/16'
```
eriqua, line 18: I'd go with a /26 max, 64 IPs is already enough for a test

modules/Microsoft.Sql/managedInstances/.test/common/dependencies.bicep
AlexanderSehr: resolved as per our discussion

modules/Microsoft.AAD/DomainServices/.test/common/dependencies.bicep
AlexanderSehr: But does that matter? Should we use a different cidr for each test - i.e. the one that is needed? Personally I'd prefer to keep the address spaces somewhat aligned as it makes seach & replace easier (if need be). 

Currently, aside from tests with multiple subnets, we either use /24 or /16.

eriqua: It doesn't matter to specify for each test a different one or the one needed. I'd at least avoid to use a /16. I think /26 would be enough for all tests

ahmadabdalla: I'd suggest to keep it as /16. There is no harm to have a bigger one. Azure resources subnet recommendations are not standardised. Example: App GW recommendation is a /24, Vwan Hub /24, azure Vmware /22. ASE I think is also /24.

eriqua (approved):
🆗

modules/Microsoft.AAD/DomainServices/.test/common/dependencies.bicep
eriqua: Accepted 😜

PR #2402, opened 2022-12-02
[Modules] Align CMK implementation for Cognitive Services & cleaned up several allowed values

AlexanderSehr:
# Description

- Align CMK implementation for Cognitive Services
- Cleaned up redundant 'allowedValues' for boolean properties from serveral modules

| Pipeline |
| - |

# Type of Change

- [ ] Update to documentation


modules/Microsoft.CognitiveServices/accounts/.test/common/deploy.test.bicep
```diff
@@ -63,6 +63,7 @@ module testDeployment '../../deploy.bicep' = {
     kind: 'Face'
     customSubDomainName: '<<namePrefix>>xdomain'
     diagnosticLogsRetentionInDays: 7
+    enableEncryption: false
```
eriqua, line 66: So do I get this right that the implementation is now aligned but we still miss prerequisites to test it in our own environment? Would that be the case also for the MS managed keys encryption?

AlexanderSehr: No-ish.
When I enable encryption without a CMK I actualy do get an error along the lines of 'this not not yet available' or so. However, when I check the [docs](https://learn.microsoft.com/en-us/azure/cognitive-services/computer-vision/identity-encrypt-data-at-rest#about-cognitive-services-encryption) it reads like Service-Managed encryption is anyways enabled.

So what I did now was to align it even closer to the other implementation in that you only have the choice of CMK - or not. Without any additional implementation for Service-Managed keys.

While cleaning up the boolean I noticed also that some of our implementation use 'allowedValues' for a bool property and removed those too. It's just very few files fortunately.

modules/Microsoft.CognitiveServices/accounts/deploy.bicep
```diff
@@ -195,7 +207,22 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
   }
 }
 
-resource cognitiveServices 'Microsoft.CognitiveServices/accounts@2021-10-01' = {
+resource cmkKeyVault 'Microsoft.KeyVault/vaults@2021-06-01-preview' existing = if (!empty(cMKKeyVaultResourceId)) {
```
eriqua, line 210: Let's align on the case
```suggestion
resource cMKKeyVault 'Microsoft.KeyVault/vaults@2021-06-01-preview' existing = if (!empty(cMKKeyVaultResourceId)) {
```

modules/Microsoft.CognitiveServices/accounts/deploy.bicep
```diff
@@ -195,7 +207,22 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
...
+  name: last(split(cMKKeyVaultResourceId, '/'))
+  scope: resourceGroup(split(cMKKeyVaultResourceId, '/')[2], split(cMKKeyVaultResourceId, '/')[4])
+}
+
+resource cMKKeyVaultKey 'Microsoft.KeyVault/vaults/keys@2021-10-01' existing = if (!empty(cMKKeyVaultResourceId) && !empty(cMKKeyName)) {
+  name: '${last(split(cMKKeyVaultResourceId, '/'))}/${cMKKeyName}'
+  scope: resourceGroup(split(cMKKeyVaultResourceId, '/')[2], split(cMKKeyVaultResourceId, '/')[4])
+}
+
+resource cMkUserAssignedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' existing = if (!empty(cMKUserAssignedIdentityResourceId)) {
```
eriqua, line 220: Let's align on the case
```suggestion
resource cMKUserAssignedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' existing = if (!empty(cMKUserAssignedIdentityResourceId)) {
```

modules/Microsoft.CognitiveServices/accounts/deploy.bicep
```diff
@@ -215,11 +242,20 @@ resource cognitiveServices 'Microsoft.CognitiveServices/accounts@2021-10-01' = {
     allowedFqdnList: allowedFqdnList
     apiProperties: apiProperties
     disableLocalAuth: disableLocalAuth
-    encryption: !empty(encryption) ? encryption : null
+    encryption: !empty(cMKKeyName) ? {
```
eriqua, line 245: If  `cMKKeyName` makes all other `keyVaultProperties` required, then I guess all other `keyVaultProperties` should also have a `Conditional` parameter category.

AlexanderSehr: Updated whereever applicable

eriqua: Although out of the scope of this PR, but since we started, I think we need to align the implementation and parameter category of `cMKUserAssignedIdentityResourceId` as well where applicable. 

ML workspaces and ServiceBus use `cMKUserAssignedIdentityResourceId`, but implemented like 
```
identity: !empty(cMKUserAssignedIdentityResourceId) ? {
        userAssignedIdentity: cMKUserAssignedIdentityResourceId
      } : null
```
which correctly makes the parameter `Optional`. 
- If the same implementation makes sense also for automation account and cognitive services, then I'd suggest to align that, updating the identity implementation and changing `cMKUserAssignedIdentityResourceId` from `Conditional` to `Optional`
- otherwise, we should update the `cMKUserAssignedIdentityResourceId` category from `Optional` to `Conditional` in the automation account module, since its current implementation depends on `cMKKeyName`.

AlexanderSehr: It is in fact different from resource to resource. Some enforce User-Assigned-Identities, some allow System-Assigned-Identities, etc.

I would highly recommend to keep this out of this PR. If we don't trust the implementations that are there now, we should revist all components and test again the difference scenarios.

AlexanderSehr: I do agree however, that it should match the current implementation. So in your example above we should use 'Optional' - not 'Conditional'

AlexanderSehr: Here it seems to be required

AlexanderSehr: addressed

PR #2424, opened 2022-12-09
[Modules] Added property `anonymousPullEnabled`

AlexanderSehr:
# Description

- Added property `anonymousPullEnabled`

| Pipeline |
| - |
| [ContainerRegistry: Registries](https://github.com/Azure/ResourceModules/actions/workflows/ms.containerregistry.registries.yml) |

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/modules/Microsoft.ContainerRegistry/registries/deploy.bicep
+++ b/modules/Microsoft.ContainerRegistry/registries/deploy.bicep
@@ -170,6 +170,9 @@ param diagnosticEventHubName string = ''
 @description('Optional. The name of the diagnostic setting, if deployed.')
 param diagnosticSettingsName string = '${name}-diagnosticSettings'
 
+@description('Optional. Enables registry-wide pull from unauthenticated clients. It\'s in preview and available in the Standard and Premium service tiers.')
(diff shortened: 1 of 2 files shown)


modules/Microsoft.ContainerRegistry/registries/deploy.bicep
```diff
@@ -170,6 +170,9 @@ param diagnosticEventHubName string = ''
 @description('Optional. The name of the diagnostic setting, if deployed.')
 param diagnosticSettingsName string = '${name}-diagnosticSettings'
 
+@description('Optional. Enables registry-wide pull from unauthenticated clients. It\'s in preview and available in the Standard and Premium service tiers.')
```
eriqua, line 173: Should this be tested as enabled in the common variant? Also, what should the default be?

AlexanderSehr: The default should be false, as it should pretty much always be for anonymous access.
I didn't add it to the test simply because it is a boolean with a value -  +so the only thing the test would do is to test if Azure works, not the template

PR #2488, opened 2022-12-21
[Modules]: Recovery Services - Updated API Versions to latest supported GA Version

ChrisSidebotham:
# Description

Updated API versions of modules under Microsoft.RecoveryServices to the latest GA version to fix failing pester tests

| Pipeline |
| - |
| [RecoveryServices: Vaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.recoveryservices.vaults.yml) |

# Type of Change

- [ ] Update to documentation

# Checklist

- [x] I did format my code

### DIFF
--- a/modules/Microsoft.RecoveryServices/vaults/deploy.bicep
+++ b/modules/Microsoft.RecoveryServices/vaults/deploy.bicep
@@ -161,7 +161,7 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
   }
 }
 
-resource rsv 'Microsoft.RecoveryServices/vaults@2022-09-10' = {
+resource rsv 'Microsoft.RecoveryServices/vaults@2022-10-01' = {
   name: name
   location: location
   tags: tags
--- a/modules/Microsoft.RecoveryServices/vaults/backupConfig/deploy.bicep
+++ b/modules/Microsoft.RecoveryServices/vaults/backupConfig/deploy.bicep
@@ -68,7 +68,7 @@ resource rsv 'Microsoft.RecoveryServices/vaults@2022-09-10' existing = {
   name: recoveryVaultName
 }
 
-resource backupConfig 'Microsoft.RecoveryServices/vaults/backupconfig@2022-02-01' = {
+resource backupConfig 'Microsoft.RecoveryServices/vaults/backupconfig@2022-04-01' = {
   name: name
   parent: rsv
   properties: {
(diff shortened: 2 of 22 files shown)


modules/Microsoft.RecoveryServices/vaults/deploy.bicep
```diff
@@ -173,6 +176,7 @@ resource rsv 'Microsoft.RecoveryServices/vaults@2022-09-10' = {
   properties: {
     monitoringSettings: !empty(monitoringSettings) ? monitoringSettings : null
     securitySettings: !empty(securitySettings) ? securitySettings : null
+    publicNetworkAccess: enablePublicNetworkAccess ? 'Enabled' : 'Disabled'
```
eriqua, line 179: For this I'd suggest to follow what done for other modules: disable publicNetworkAccess by default if private endpoint is set. This would also involve to add an empty string as an allowed value for the publicNetworkAccess parameter. You can reference the implementation in the ML workspace module https://github.com/Azure/ResourceModules/blob/main/modules/Microsoft.DataFactory/factories/deploy.bicep#L199

ChrisSidebotham: Updated in latest PR - PublicnetworkAccess now using `!empty(publicNetworkAccess) ? any(publicNetworkAccess) : (!empty(privateEndpoints) ? 'Disabled' : null)`

modules/Microsoft.RecoveryServices/vaults/.test/common/deploy.test.bicep
```diff
@@ -58,6 +58,7 @@ module testDeployment '../../deploy.bicep' = {
   scope: resourceGroup
   name: '${uniqueString(deployment().name)}-test-${serviceShort}'
   params: {
+    publicNetworkAccess: 'Enabled'
```
eriqua, line 61: Since the updated implementation, we should remove this setting. `publicNetworkAccess` will be disabled by default due to the implementation of private endpoints 
```suggestion
```

modules/Microsoft.Compute/virtualMachines/deploy.bicep
```diff
@@ -466,6 +466,7 @@ resource vm 'Microsoft.Compute/virtualMachines@2021-07-01' = {
           deleteOption: contains(nicConfiguration, 'deleteOption') ? nicConfiguration.deleteOption : 'Delete'
           primary: index == 0 ? true : false
         }
+        #disable-next-line use-resource-id-functions
```
eriqua, line 469: There's something we need to double check about the implementation of networkProfile in VM vs the one in VMSS. In VMSS we implement that differently and we don't have this warning. I'd suggest to remove this workaround to have this failing and try to solve it instead.
```suggestion
```

eriqua (changes requested):
Great job as always! Could you just please revert the changes to the utilities/src/apiSpecsList.json file? Automatic changes to this file were due to a bug which should now be solved

eriqua (approved):
🆗

PR #2515, opened 2022-12-29
[Modules] Azure Virtual Network Manager Module (API Version: 2022-07-01)

### LINKED ISSUE #2419: [Feature Request]: Azure Virtual Network Manager (AVNM) Module
### Description

Review and add the AVNM resource into our library. It is still in public preview. Needs further investigations on testing and deployment in the CARML context.

https://learn.microsoft.com/en-us/azure/templates/microsoft.network/networkmanagers?pivots=deployment-language-bicep

ahmadabdalla:
# Description

Closes #2419  

This PR adds the `Microsoft.Network/networkManagers` module. It uses the currently published API Version `2022-07-01`. 

---
**Note**: There is one missing child resource called `securityUserConfigurations` and its nested children modules. This resource type is a new feature that is not yet available for deployments or testing although the template and API documentation reference it and is well documented. Attempting to test it fails. Network Manager module will be need to be updated in the future to support it.

---

## Module Parents and Children
- networkManagers [Parent]
  - Network Groups [Child]
    - Static Members [Child]
  - Connectivity Configurations [Child]
  - Scope Connections
  - Security Admin Configurations
    - Rule Collections [Child]
      - Rules [Child]

## Test Cases
 - **Min**: Only the `name` parameter.
 - **Common**: Tests all parameter combinations, with child modules.

## Extension Resources
 - Role Assignments: Updated to cover roles needed for the Network Manager Lab resource type.

## Pipeline references

As this is a new module pipeline, created in the Fork and referenced a deployment from the fork.

| Pipeline |
| - |
|[Network: Network Managers](https://github.com/ahmadabdalla/ResourceModules/actions/workflows/ms.network.networkmanagers.yml)|

# Type of Change

# Checklist

- [x] I did format my code


ahmadabdalla:
Note:

In the [common deploy.test.bicep file](https://github.com/ahmadabdalla/ResourceModules/blob/users/ahmad/2419_avnm_carml_aa/modules/Microsoft.Network/networkManagers/.test/common/deploy.test.bicep), there is a reference to an existing `NetworkGroup` by using the `ResourceID` format like the below:
```bicep
networkGroupId: concat(resourceGroup.id, '/providers/Microsoft.Network/networkManagers/,<<namePrefix>>${serviceShort}001,/networkGroups/network-group-spokes')
```

The reason it is implemented this way is that as per the API and template [documentation](https://learn.microsoft.com/en-us/azure/templates/microsoft.network/networkmanagers/connectivityconfigurations?pivots=deployment-language-bicep#connectivitygroupitem), the `NetworkGroup` is not referenced using the name, but uses the resource ID. This means that we cannot interpret the Resource ID before hand given we are deploying this child resource. Hence I have constructed it manually and created the dependency in the main module as the dependency is valid if someone wants to deploy the whole NVM in a single deployment template, where Network Groups must be created before configurations. 

However, I would like feedback on the current approach if we need to pivot. Example: Another option is to change the actual input to Name and then create an existing function in the other child modules to evaluate the ID, but I did not want to introduce this change unless we discuss the current approach.

modules/Microsoft.Network/networkManagers/.test/common/deploy.test.bicep
```diff
@@ -0,0 +1,203 @@
...
+            isGlobal: 'False'
+          }
+        ]
+      }
+    ]
+    scopeConnections: [
+      {
+        name: 'scope-connection-test'
+        description: 'description of the scope connection'
+        resourceId: '/subscriptions/<<subscriptionId>>'
```
AlexanderSehr, line 129: Should work via the `subscription().id` function

modules/Microsoft.Network/networkManagers/.test/min/deploy.test.bicep
```diff
@@ -0,0 +1,49 @@
...
+  name: '${uniqueString(deployment().name)}-test-${serviceShort}'
+  params: {
+    enableDefaultTelemetry: enableDefaultTelemetry
+    name: '<<namePrefix>>${serviceShort}001'
+    networkManagerScopeAccesses: [
+      'Connectivity'
+    ]
+    networkManagerScopes: {
+      managementGroups: [
+        '/providers/Microsoft.Management/managementGroups/<<managementGroupId>>'
```
AlexanderSehr, line 45: Can this be done via the `managementGroup()` function

modules/Microsoft.Network/networkManagers/connectivityConfigurations/deploy.bicep
```diff
@@ -0,0 +1,78 @@
...
+
+@sys.description('Conditional. List of hub items. This will create peerings between the specified hub and the virtual networks in the network group specified. Required if connectivityTopology is of type "HubAndSpoke".')
+param hubs array = []
+
+@allowed([
+  'True'
+  'False'
+])
+@sys.description('Optional. Flag if need to remove current existing peerings. If set to "True", all peerings on virtual networks in selected network groups will be removed and replaced with the peerings defined by this configuration. Optional when connectivityTopology is of type "HubAndSpoke". Default is "False".')
+param deleteExistingPeering string = 'False'
```
AlexanderSehr, line 30: Should we treat this as a boolean?

AlexanderSehr: Same for the other parameters

modules/Microsoft.Network/networkManagers/networkGroups/readme.md
```diff
@@ -0,0 +1,57 @@
...
+| Parameter Name | Type | Default Value | Description |
+| :-- | :-- | :-- | :-- |
+| `description` | string | `''` | A description of the network group. |
+| `enableDefaultTelemetry` | bool | `True` | Enable telemetry via a Globally Unique Identifier (GUID). |
+| `staticMembers` | _[staticMembers](staticMembers/readme.md)_ array | `[]` | Static Members to create for the network group. Contains virtual networks to add to the network group. |
+
+
+### Parameter Usage: `<ParameterPlaceholder>`
+
+// TODO: Fill in Parameter usage
```
AlexanderSehr, line 45: #placeholder

modules/Microsoft.Network/networkManagers/networkGroups/staticMembers/readme.md
```diff
@@ -0,0 +1,56 @@
...
+**Optional parameters**
+
+| Parameter Name | Type | Default Value | Description |
+| :-- | :-- | :-- | :-- |
+| `enableDefaultTelemetry` | bool | `True` | Enable telemetry via a Globally Unique Identifier (GUID). |
+
+
+### Parameter Usage: `<ParameterPlaceholder>`
+
+// TODO: Fill in Parameter usage
```
AlexanderSehr, line 44: #placeholder

modules/Microsoft.Network/networkManagers/readme.md
```diff
@@ -0,0 +1,608 @@
...
+| `lock` | string | `''` | `['', CanNotDelete, ReadOnly]` | Specify the type of lock. |
+| `roleAssignments` | array | `[]` |  | Array of role assignment objects that contain the 'roleDefinitionIdOrName' and 'principalId' to define RBAC role assignments on this resource. In the roleDefinitionIdOrName attribute, you can provide either the display name of the role definition, or its fully qualified ID in the following format: '/providers/Microsoft.Authorization/roleDefinitions/c2f4ef07-c644-48eb-af81-4b1b4947fb11'. |
+| `scopeConnections` | _[scopeConnections](scopeConnections/readme.md)_ array | `[]` |  | Scope Connections to create for the network manager. Allows network manager to manage resources from another tenant. |
+| `securityAdminConfigurations` | _[securityAdminConfigurations](securityAdminConfigurations/readme.md)_ array | `[]` |  | Security Admin Configurations, Rule Collections and Rules to create for the network manager. |
+| `tags` | object | `{object}` |  | Tags of the resource. |
+
+
+### Parameter Usage: `<ParameterPlaceholder>`
+
+// TODO: Fill in Parameter usage
```
AlexanderSehr, line 62: #placeholder

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/deploy.bicep
```diff
@@ -0,0 +1,67 @@
...
+@maxLength(64)
+@sys.description('Required. The name of the security admin configuration.')
+param name string
+
+@maxLength(500)
+@sys.description('Optional. A description of the security admin configuration.')
+param description string = ''
+
+@sys.description('Required. Enum list of network intent policy based services. String array containing any of: "All", "AllowRulesOnly" or "None". Default is "None".')
+param applyOnNetworkIntentPolicyBasedServices array = [ 'None' ]
```
AlexanderSehr, line 13: Instead of having it in the description I'd recommend an allowed set

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/readme.md
```diff
@@ -0,0 +1,59 @@
...
+| Parameter Name | Type | Default Value | Description |
+| :-- | :-- | :-- | :-- |
+| `description` | string | `''` | A description of the security admin configuration. |
+| `enableDefaultTelemetry` | bool | `True` | Enable telemetry via a Globally Unique Identifier (GUID). |
+| `ruleCollections` | _[ruleCollections](ruleCollections/readme.md)_ array | `[]` | A security admin configuration contains a set of rule collections that are applied to network groups. Each rule collection contains one or more security admin rules. |
+
+
+### Parameter Usage: `<ParameterPlaceholder>`
+
+// TODO: Fill in Parameter usage
```
AlexanderSehr, line 47: #placeholder

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/ruleCollections/readme.md
```diff
@@ -0,0 +1,59 @@
...
+| Parameter Name | Type | Default Value | Description |
+| :-- | :-- | :-- | :-- |
+| `description` | string | `''` | A description of the admin rule collection. |
+| `enableDefaultTelemetry` | bool | `True` | Enable telemetry via a Globally Unique Identifier (GUID). |
+| `rules` | _[rules](rules/readme.md)_ array |  | List of rules for the admin rules collection. Security admin rules allows enforcing security policy criteria that matches the conditions set. Warning: A rule collection without rule will cause a deployment configuration for security admin goal state in network manager to fail. |
+
+
+### Parameter Usage: `<ParameterPlaceholder>`
+
+// TODO: Fill in Parameter usage
```
AlexanderSehr, line 47: #placeholder

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/ruleCollections/rules/readme.md
```diff
@@ -0,0 +1,67 @@
...
+| `enableDefaultTelemetry` | bool | `True` |  | Enable telemetry via a Globally Unique Identifier (GUID). |
+| `sourcePortRanges` | array | `[]` |  | List of destination port ranges. This specifies on which ports traffic will be allowed or denied by this rule. Provide an (*) to allow traffic on any port. Port ranges are between 1-65535. |
+| `sourcesAddressPrefix` | string | `''` |  | Provide an address range using CIDR notation (e.g. 192.168.99.0/24 or 2001:1234::/64), or an IP address (e.g. 192.168.99.0 or 2001:1234::), or a service tag (e.g. AppService.WestEurope). You can also provide a comma-separated list of IP addresses or address ranges using either IPv4 or IPv6. |
+| `sourcesAddressPrefixType` | string | `''` | `['', IPPrefix, ServiceTag]` | The source filter can be an IP Address or a service tag. It specifies the incoming traffic from a specific source IP addresses range that will be allowed or denied by this rule. |
+
+
+### Parameter Usage: `<ParameterPlaceholder>`
+
+// TODO: Fill in Parameter usage
```
AlexanderSehr, line 55: #placeholder

modules/Microsoft.Network/networkManagers/.test/common/deploy.test.bicep
ahmadabdalla: I left this on purpose to demonstrate that it can be multiple values. I'll update it and provide details in a parameter usage section.

modules/Microsoft.Network/networkManagers/.test/min/deploy.test.bicep
ahmadabdalla: We'd need to change the scope to an MG scope deployment to use that function. Right?

modules/Microsoft.Network/networkManagers/connectivityConfigurations/deploy.bicep
ahmadabdalla: Aligning to the template documentation and the API reference - It's a string. I would prefer to keep things as close as possible to avoid confusion.

modules/Microsoft.Network/networkManagers/networkGroups/readme.md
ahmadabdalla: i thought I removed those 🤦🏻‍♂️

modules/Microsoft.Network/networkManagers/.test/common/deploy.test.bicep
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/networkGroups/readme.md
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/networkGroups/staticMembers/readme.md
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/readme.md
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/scopeConnections/readme.md
```diff
@@ -0,0 +1,57 @@
...
+
+| Parameter Name | Type | Default Value | Description |
+| :-- | :-- | :-- | :-- |
+| `description` | string | `''` | A description of the scope connection. |
+| `enableDefaultTelemetry` | bool | `True` | Enable telemetry via a Globally Unique Identifier (GUID). |
+
+
+### Parameter Usage: `<ParameterPlaceholder>`
+
+// TODO: Fill in Parameter usage
```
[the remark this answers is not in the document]
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/readme.md
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/ruleCollections/readme.md
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/ruleCollections/rules/readme.md
ahmadabdalla: @AlexanderSehr Updated

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/deploy.bicep
ahmadabdalla: Agreed. @AlexanderSehr Updated.

modules/Microsoft.Network/networkManagers/.test/min/deploy.test.bicep
AlexanderSehr: If so, nevermind 😄 will check later. You might be right that this function behaves a bit weird (for no reason)

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/deploy.bicep
AlexanderSehr: Thanks. Now let's also remove the default in the description 😄

ahmadabdalla: done :)

modules/Microsoft.Network/networkManagers/.test/min/deploy.test.bicep
ahmadabdalla: It is indeed based on the doco and what I've tested too:

See [Documentation](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions-scope#managementgroup)

_managementGroup() can only be used on a [management group deployments](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/deploy-to-management-group). It returns the current management group for the deployment operation. Use when either getting a scope object or getting properties for the current management group._

AlexanderSehr: A shame. One would think ARM would be able to detect the ManagementGroup of a given subscription

modules/Microsoft.Network/networkManagers/connectivityConfigurations/deploy.bicep
AlexanderSehr: Hmm, this is indeed quite unfortunate. It just seems to unnecessary 'wrong' to me. I wonder why the RP did not implement a boolean when they had the chance. Also the name does not indicate that there'll ever be another value aside from 'true' & 'false'. If we however want to keep the string, I'd vouch to at least add an allowedSet, as silly as it seems.

@eriqua any thoughts from your end?

AlexanderSehr:
> Note:
> 
> In the [common deploy.test.bicep file](https://github.com/ahmadabdalla/ResourceModules/blob/users/ahmad/2419_avnm_carml_aa/modules/Microsoft.Network/networkManagers/.test/common/deploy.test.bicep), there is a reference to an existing `NetworkGroup` by using the `ResourceID` format like the below:
> 
> ```bicep
> networkGroupId: concat(resourceGroup.id, '/providers/Microsoft.Network/networkManagers/,<<namePrefix>>${serviceShort}001,/networkGroups/network-group-spokes')
> ```
> 
> The reason it is implemented this way is that as per the API and template [documentation](https://learn.microsoft.com/en-us/azure/templates/microsoft.network/networkmanagers/connectivityconfigurations?pivots=deployment-language-bicep#connectivitygroupitem), the `NetworkGroup` is not referenced using the name, but uses the resource ID. This means that we cannot interpret the Resource ID before hand given we are deploying this child resource. Hence I have constructed it manually and created the dependency in the main module as the dependency is valid if someone wants to deploy the whole NVM in a single deployment template, where Network Groups must be created before configurations.
> 
> However, I would like feedback on the current approach if we need to pivot. Example: Another option is to change the actual input to Name and then create an existing function in the other child modules to evaluate the ID, but I did not want to introduce this change unless we discuss the current approach.

Hey @ahmadabdalla, we do in fact have similar 'odd' cases. For example the [AppGW](https://github.com/Azure/ResourceModules/blob/main/modules/Microsoft.Network/applicationGateways/.test/common/deploy.test.bicep#:~:text=var%20appGWExpectedResourceID%20%3D%20%27%24%7BresourceGroup.id%7D/providers/Microsoft.Network/applicationGateways/%24%7BappGWName%7D%27) also has a child that wants to reference resource IDs of its linked resources. There, we introduced a variable `var appGWExpectedResourceID = '${resourceGroup.id}/providers/Microsoft.Network/applicationGateways/${appGWName}'` which we then use e.g. like `id: '${appGWExpectedResourceID}/frontendIPConfigurations/public'`. It's in any case quite - strange.

modules/Microsoft.Network/networkManagers/connectivityConfigurations/deploy.bicep
ahmadabdalla: > Hey @ahmadabdalla, we do in fact have similar 'odd' cases. For example the [AppGW](https://github.com/Azure/ResourceModules/blob/main/modules/Microsoft.Network/applicationGateways/.test/common/deploy.test.bicep#:~:text=var%20appGWExpectedResourceID%20%3D%20%27%24%7BresourceGroup.id%7D/providers/Microsoft.Network/applicationGateways/%24%7BappGWName%7D%27) also has a child that wants to reference resource IDs of its linked resources. There, we introduced a variable `var appGWExpectedResourceID = '${resourceGroup.id}/providers/Microsoft.Network/applicationGateways/${appGWName}'` which we then use e.g. like `id: '${appGWExpectedResourceID}/frontendIPConfigurations/public'`. It's in any case quite - strange.

@AlexanderSehr will align to the approach above. At least we're in an agreement in principle :)

ahmadabdalla:
> Hey @ahmadabdalla, we do in fact have similar 'odd' cases. For example the [AppGW](https://github.com/Azure/ResourceModules/blob/main/modules/Microsoft.Network/applicationGateways/.test/common/deploy.test.bicep#:~:text=var%20appGWExpectedResourceID%20%3D%20%27%24%7BresourceGroup.id%7D/providers/Microsoft.Network/applicationGateways/%24%7BappGWName%7D%27) also has a child that wants to reference resource IDs of its linked resources. There, we introduced a variable `var appGWExpectedResourceID = '${resourceGroup.id}/providers/Microsoft.Network/applicationGateways/${appGWName}'` which we then use e.g. like `id: '${appGWExpectedResourceID}/frontendIPConfigurations/public'`. It's in any case quite - strange.

@AlexanderSehr will align to the approach above. At least we're in an agreement in principle :)

modules/Microsoft.Network/networkManagers/connectivityConfigurations/deploy.bicep
ahmadabdalla: We will have to do this conversion on every parameter in this resource type as unfortunately none of them are Booleans
image

so would be something like this on each property:
```bicep
isGlobal: string(isGlobal)
```
I've already applied the 'allowed set' to each of those. @eriqua thoughts on this one please?

modules/Microsoft.Network/networkManagers/securityAdminConfigurations/ruleCollections/deploy.bicep
```diff
@@ -0,0 +1,86 @@
...
+  name: '${uniqueString(deployment().name)}-RuleCollections-Rules-${index}'
+  params: {
+    networkManagerName: networkManager.name
+    securityAdminConfigurationName: securityAdminConfigurationName
+    ruleCollectionName: ruleCollection.name
+    name: rule.name
+    access: rule.access
+    description: contains(rule, 'description') ? rule.description : ''
+    destinationPortRanges: contains(rule, 'destinationPortRanges') ? rule.destinationPortRanges : []
+    destinations: contains(rule, 'destinations') ? rule.destinations : []
```
ahmadabdalla, line 65: @AlexanderSehr I have added the following new properties `destinations` and `sources` to handle a use case where multiple IPs are required for either a source or a destination. 

**BEFORE**
```bicep
destinations: !empty(destinationsAddressPrefix) && !empty(destinationsAddressPrefixType) ? [
      {
        addressPrefix: destinationsAddressPrefix // this value is defined as a parameter
        addressPrefixType: destinationsAddressPrefixType // this value is defined as a parameter
      }
    ] : [ ] 
```

**AFTER**

```bicep
    destinations: !empty(destinationsAddressPrefix) && !empty(destinationsAddressPrefixType) && empty(destinations) ? [ // added another condition
      {
        addressPrefix: destinationsAddressPrefix
        addressPrefixType: destinationsAddressPrefixType
      }
    ] : destinations // this is no longer passed empty here
```
-------------

Now users have the choice to use a single entry using the parameters `destinationsAddressPrefix` and `destinationsAddressPrefixType` (Method-1) 

**OR**

provide the input into the `destinations` object which can support multiple entries (Method-2):

image
In this example: **Method-2** takes precedence over **Method-1**

I wanted to highlight this change as I've introduced it after the review. Let me know your thoughts on this one. Should we:
- Allow both methods for flexibility OR
- get rid of **method-1** and only support **method-2**

Method-2 is how we always do things for child properties. But I thought of introducing some flexibility.

AlexanderSehr: Hey @ahmadabdalla, while I like the flexibility, I wonder if it would ultimately lead to more confusion as one has to understand two approaches instead of one - and use either one based on the use case.
To this end I'm inclined to vote for option 2, that is, `get rid of method-1 and only support method-2`. But I'll leave it up to you :D

ahmadabdalla: Thanks @AlexanderSehr After giving it some thought, I've decided to stick to `method-2` and remove support for `method-1`. Will be resolving this comment as well. Based on the outcome of this discussion.

modules/Microsoft.Network/networkManagers/connectivityConfigurations/deploy.bicep
ahmadabdalla: @AlexanderSehr @eriqua Shall we resolve this comment? Its the only one left outstanding if there is no further feedback. Current implementation applies the allow set concept to restrict usage of the parameters.

AlexanderSehr: @eriqua is out of office this week, so let's give her a bit more time :)

AlexanderSehr: Sounds good to me. @ahmadabdalla? If you're fine with that we can resolve this comment and create the issue @eriqua proposed

ahmadabdalla: Happy with the suggestion. @eriqua @AlexanderSehr I've created an issue for it #2546.

AlexanderSehr: Great, thanks 😊

AlexanderSehr (approved):

PR #2530, opened 2023-01-05
[Modules]: Microsoft.RecoveryServices - Param and Readme fix failing pipeline

ChrisSidebotham:
# Description

Updating readme file and publicNetworkAccess default values following PR #2488 to fix failing deployment

| Pipeline |
| - |
| [RecoveryServices: Vaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.recoveryservices.vaults.yml)|

# Type of Change

- [ ] Update to documentation

# Checklist

- [x] I did format my code

### DIFF
--- a/modules/Microsoft.RecoveryServices/vaults/deploy.bicep
+++ b/modules/Microsoft.RecoveryServices/vaults/deploy.bicep
@@ -116,7 +116,7 @@ param securitySettings object = {}
   'Enabled'
   'Disabled'
 ])
-param publicNetworkAccess string = ''
+param publicNetworkAccess string = 'Disabled'
 
 var diagnosticsLogsSpecified = [for category in filter(diagnosticLogCategoriesToEnable, item => item != 'allLogs'): {
   category: category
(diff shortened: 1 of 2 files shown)


modules/Microsoft.RecoveryServices/vaults/deploy.bicep
```diff
@@ -112,11 +112,10 @@ param securitySettings object = {}
 
 @description('Optional. Whether or not public network access is allowed for this resource. For security reasons it should be disabled. If not specified, it will be disabled by default if private endpoints are set.')
 @allowed([
-  ''
   'Enabled'
   'Disabled'
 ])
-param publicNetworkAccess string = ''
+param publicNetworkAccess string = 'Disabled'
```
AlexanderSehr, line 118: Up until now we used `null` as a default value if nothing was provided (including no Private Endpoints). So I take it that `null` isn't a valid value anymore?

If that's the case, I think we should also simplify the logic `publicNetworkAccess: !empty(publicNetworkAccess) ? any(publicNetworkAccess) : (!empty(privateEndpoints) ? 'Disabled' : null)` itself, as the parameter cannot be `empty` anymore.

I think so far we used `''` as a default as we figured a user wouldn't need to provide the value - and only if he/she did we'd use it. Otherwise we'd assume a value based on whether a private endpoint was provided or not.

If we change this because it isn't working anymore, also the parameter's description must be updated.

ChrisSidebotham: Changes committed as addressed in the review. Changing to assign param instead of logic due to the unsupported nature of `null` for the mapped NetworkSettings param during deployment

AlexanderSehr (approved):

PR #2560, opened 2023-01-13
[Wiki] Added initial version of troubleshooting guidance

AlexanderSehr:
# Description

- Added initial version of troubleshooting guidance

# Type of Change

- [x] Update to documentation


docs/wiki/The CI environment - Troubleshooting.md
```diff
@@ -0,0 +1,35 @@
...
+
+    instead of
+
+    ```json
+    {
+      "it's a me"
+    }
+    ```
+
+  - **Background:** If configured as a multiline object, GitHub interprets each line as a dedicated secret value. This leads to a situation where characters like `{` & `}` are treated as individual secrets - and values that are passed as JSON strings between pipeline jobs that contain these characters cannot be passed anymore (as per GitHubs default behavior). For example: `{"removeDeployment": true}`. Because this value isn't passed anymore, subsequently the line `removeDeployment: '${{ (fromJson(needs.job_initialize_pipeline.outputs.workflowInput)).removeDeployment }}'` that references is throws the error message you're seeing.
```
eriqua, line 35:
```suggestion
  - **Background:** If configured as a multiline object, GitHub interprets each line as a dedicated secret value. This leads to a situation where characters like `{` & `}` are treated as individual secrets - and values that are passed as JSON strings between pipeline jobs that contain these characters cannot be passed anymore (as per GitHubs default behavior). For example: `{"removeDeployment": true}`. Because this value isn't passed anymore, subsequently the line `removeDeployment: '${{ (fromJson(needs.job_initialize_pipeline.outputs.workflowInput)).removeDeployment }}'` that references it throws the error message you're seeing.
```

eriqua (approved):
🆗

PR #2636, opened 2023-01-31
[Modules] Updated Disk Encryption Set to latest published API version (2022-07-02)

### LINKED ISSUE #2632: [Failed pipeline] Compute: DiskEncryptionSets
failed run: https://github.com/Azure/ResourceModules/actions/runs/4048059986

ahmadabdalla:
Closes #2632 

# Description

Updated the DES module to the latest API version to fix failing pipeline: 

Added support for:
- federatedClientId: Multi-tenant application client id to access key vault in a different tenant. 
- User Assigned Managed Identity support. Set the system assigned to default to true.

| Pipeline |
| - |
|[Compute: DiskEncryptionSets](https://github.com/Azure/ResourceModules/actions/workflows/ms.compute.diskencryptionsets.yml) |

# Type of Change

# Checklist

- [ ] I did format my code

### DIFF
--- a/modules/Microsoft.Compute/diskEncryptionSets/.test/common/dependencies.bicep
+++ b/modules/Microsoft.Compute/diskEncryptionSets/.test/common/dependencies.bicep
@@ -7,6 +7,11 @@ param keyVaultName string
 @description('Required. The name of the Managed Identity to create.')
 param managedIdentityName string
 
+resource managedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' = {
+    name: managedIdentityName
+    location: location
+}
+
 resource keyVault 'Microsoft.KeyVault/vaults@2022-07-01' = {
     name: keyVaultName
     location: location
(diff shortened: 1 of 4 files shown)


modules/Microsoft.Compute/diskEncryptionSets/.test/common/dependencies.bicep
```diff
@@ -16,7 +16,7 @@ resource keyVault 'Microsoft.KeyVault/vaults@2022-07-01' = {
             name: 'standard'
         }
         tenantId: tenant().tenantId
-        enablePurgeProtection: null
+        enablePurgeProtection: true
```
AlexanderSehr, line 19: Does it have to be true? If not I strongly recommend to set it to `false`. Not that we don't generate a new name each time we run. As you ran a test case there's now already a fair chance we cannot re-run the test for the next couple days :D

ahmadabdalla: Hmmm.. it was complaining it needed to be true unfortunately.. given it was working before, maybe the new API mandates this.. but I was blocked by the purge protection on KV must be set to true

AlexanderSehr: That's alright (might be new) - but then we must adopt the KeyVault deployment approach we use for Purge Protected Key Vaults like in the Batch Module's Encryption test

ahmadabdalla: No problem. Aligned it to the batch module for baseTime usage.

AlexanderSehr (approved):

PR #2657, opened 2023-02-06
[Modules] Added Database for MySQL Flexible Server Module

### LINKED ISSUE #2647: [Feature Request]: Add module for Azure Database for MySQL Flexible Server
### Description

Add a module for provisioning Azure Database for MySQL Flexible Server (Microsoft.DBforMySQL/flexibleServers)

ahmadabdalla:
Closes #2647 

# Description

New Module for Microsoft Database for MySQL Flexible Servers

- Supports Public and Private Servers
- Geo backed up Servers with CMK
- Child modules for (config, firewall rules and databases)

Aligns to the PostgreSQL Flexible server's module we have in the CARML library.

## Pipeline references

| Pipeline |
| - |
| [DbForMySQL: FlexibleServers](https://github.com/ahmadabdalla/ResourceModules/actions/workflows/ms.dbformysql.flexibleservers.yml)|

# Type of Change

# Checklist

- [ ] I did format my code


modules/Microsoft.DBforMySQL/flexibleServers/.bicep/nested_roleAssignments.bicep
```diff
@@ -0,0 +1,68 @@
...
+  'Monitoring Reader': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '43d0d8ad-25c7-4714-9337-8ba259a9fe05')
+  Owner: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '8e3af657-a8ff-443c-a75c-2fe8c4bcb635')
+  Reader: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'acdd72a7-3385-48ef-bd42-f606fba81ae7')
+  'Resource Policy Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '36243c78-bf99-498c-9df9-86d9f8d28608')
+  'Role Based Access Control Administrator (Preview)': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'f58310d9-a9f6-439a-9e8d-f62e7b41a168')
+  'User Access Administrator': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '18d7d88d-d35e-4fb5-a5c3-7773c20a72d9')
+}
+
+resource flexibleServer 'Microsoft.DBforMySQL/flexibleServers@2021-12-01-preview' existing = {
+  name: last(split(resourceId, '/'))
```
AlexanderSehr, line 53:
```suggestion
  name: last(split(resourceId, '/'))!
```

modules/Microsoft.DBforMySQL/flexibleServers/.test/.scripts/Get-PairedRegion.ps1
```diff
@@ -0,0 +1,16 @@
+﻿param(
+    [string] $Location
```
AlexanderSehr, line 2: Please add synopsis (like in PR #2681)

modules/Microsoft.DBforMySQL/flexibleServers/.test/.scripts/Get-PairedRegion.ps1
```diff
@@ -0,0 +1,16 @@
...
+)
+
+# Sleep for role assignment propagation
+Start-Sleep -Seconds 10
+
+$PairedRegionName = Get-AzLocation |
+    Where-Object -FilterScript { $PSItem.Location -eq $Location -or $PSItem.DisplayName -eq $Location } |
```
AlexanderSehr, line 9:
```suggestion
    Where-Object -FilterScript { $Location - in @($PSItem.Location, $PSItem.DisplayName) } |
```

ahmadabdalla: @AlexanderSehr will commit but then keep the parameter names `FilterScript` as its best practice for readability

modules/Microsoft.DBforMySQL/flexibleServers/.test/.scripts/Get-PairedRegion.ps1
AlexanderSehr: And now that the other PR (#2681) actually one more: Please move it into the shared scripts folder (`/modules/.shared/.scripts`)

ahmadabdalla: Yep I noticed that too.

modules/Microsoft.DBforMySQL/flexibleServers/deploy.bicep
```diff
@@ -0,0 +1,406 @@
...
+
+resource cMKKeyVaultKey 'Microsoft.KeyVault/vaults/keys@2022-07-01' existing = if (!empty(cMKKeyVaultResourceId) && !empty(cMKKeyName)) {
+  name: '${last(split(cMKKeyVaultResourceId, '/'))}/${cMKKeyName}'
+  scope: resourceGroup(split(cMKKeyVaultResourceId, '/')[2], split(cMKKeyVaultResourceId, '/')[4])
+}
+
+resource geoBackupCMKKeyVaultKey 'Microsoft.KeyVault/vaults/keys@2022-07-01' existing = if (!empty(geoBackupCMKKeyVaultResourceId) && !empty(geoBackupCMKKeyName)) {
+  name: '${last(split(geoBackupCMKKeyVaultResourceId, '/'))}/${geoBackupCMKKeyName}'
+  scope: resourceGroup(split(geoBackupCMKKeyVaultResourceId, '/')[2], split(geoBackupCMKKeyVaultResourceId, '/')[4])
+}
```
AlexanderSehr, line 271:
```suggestion
resource cMKKeyVault 'Microsoft.KeyVault/vaults@2022-07-01' existing = if (!empty(cMKKeyVaultResourceId)) {
  name: last(split(cMKKeyVaultResourceId, '/'))
  scope: resourceGroup(split(cMKKeyVaultResourceId, '/')[2], split(cMKKeyVaultResourceId, '/')[4])

  resource cMKKey 'keys@2022-07-01' existing = if (!empty(cMKKeyVaultResourceId) && !empty(cMKKeyName)) {
    name: cMKKeyName
  }

  resource geoBackupCMKKey 'keys@2022-07-01' existing = if (!empty(geoBackupCMKKeyVaultResourceId) && !empty(geoBackupCMKKeyName)) {
    name: geoBackupCMKKeyName
  }
}
```

This should work. Mind you though, we have to update references from for example `geoBackupCMKKeyVaultKey` to `cMKKeyVault::cMKKey`.

We may even be able to remove the 2nd and 3rd `!empty(geoBackupCMKKeyVaultResourceId)` check in the suggestion. But that must be tested as I don't know if Bicep evaluates inner resources if the condition of the outer does not apply.

modules/Microsoft.DBforMySQL/flexibleServers/.test/public/deploy.test.bicep
```diff
@@ -0,0 +1,165 @@
...
+// ============ //
+
+// General resources
+// =================
+resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
+  name: resourceGroupName
+  location: location
+}
+
+module nestedDependencies1 'dependencies1.bicep' = {
```
AlexanderSehr, line 38: Why is this split in 2? I can see that an output of the 1st dependency deployment is passed to the 2nd, but I fail do unterstand why :D

ahmadabdalla: The `location` property cannot accept an output from a previous deployment. That's why I had to split it into two, so that I can extract the paired region via the deployment script, then pass it as an input to the next dependency deployment which is in the paired region 

```bicep
This expression is being used in an assignment to the "location" property of the "Microsoft.KeyVault/vaults" type, which requires a value that can be calculated at the start of the deployment. Properties of geoBackupManagedIdentity which can be calculated at the start include "apiVersion", "id", "name", "type". Bicep(BCP120)
```

modules/Microsoft.DBforMySQL/flexibleServers/deploy.bicep
ahmadabdalla: @AlexanderSehr I've applied the suggestion. Will test things out. I 'unresolved' the issue so that I can update here

ahmadabdalla: @AlexanderSehr, thinking about this a bit more:

image

What is the value of having the condition in the inner resource? Even in the previous implementation, the `cMKKeyName` should become a conditional parameter instead of optional, as we need the name of the Key if we are using CMK. But the Key version is the one that is optional. Doesn't this also to other modules implementations we have. Example: 

```bicep
if (!empty(cMKKeyVaultResourceId) && !empty(cMKKeyName))`
```

should be

```bicep
if (!empty(cMKKeyVaultResourceId)`
```

and the deployment should fail if the `cMKKeyName` is not provided. Thoughts?

AlexanderSehr: The only reason for the inner condition is that I don't know what Bicep/ARM does if you don't provide it. For example: If no KeyVaultResourceID was provided, does that mean the inner condition isn't evaluated automatically? That would be great. But I wouldn't be surprised if it is anyways and the deployment would fail.

Regarding your point with `Conditional` for the parameter I somewhat agree. However, when viewing the `existing` resources in isolation, I guess the condtions still make sense.

Can you draft a proposal? That would make it easier to discuss :D

modules/Microsoft.DBforMySQL/flexibleServers/deploy.bicep
```diff
@@ -0,0 +1,424 @@
...
+    mode: 'Incremental'
+    template: {
+      '$schema': 'https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#'
+      contentVersion: '1.0.0.0'
+      resources: []
+    }
+  }
+}
+
+//resource cMKKeyVaultKey 'Microsoft.KeyVault/vaults/keys@2022-07-01' existing = if (!empty(cMKKeyVaultResourceId) && !empty(cMKKeyName)) {
```
ahmadabdalla, line 263: Option 1

modules/Microsoft.DBforMySQL/flexibleServers/deploy.bicep
```diff
@@ -0,0 +1,424 @@
...
+//  name: '${last(split(cMKKeyVaultResourceId, '/'))}/${cMKKeyName}'
+//  scope: resourceGroup(split(cMKKeyVaultResourceId, '/')[2], split(cMKKeyVaultResourceId, '/')[4])
+//}
+
+resource cMKKeyVault 'Microsoft.KeyVault/vaults@2022-07-01' existing = if (!empty(cMKKeyVaultResourceId)) {
```
ahmadabdalla, line 268: Option 2

modules/Microsoft.DBforMySQL/flexibleServers/deploy.bicep
```diff
@@ -0,0 +1,424 @@
...
+      resources: []
+    }
+  }
+}
+
+//resource cMKKeyVaultKey 'Microsoft.KeyVault/vaults/keys@2022-07-01' existing = if (!empty(cMKKeyVaultResourceId) && !empty(cMKKeyName)) {
+//  name: '${last(split(cMKKeyVaultResourceId, '/'))}/${cMKKeyName}'
+//  scope: resourceGroup(split(cMKKeyVaultResourceId, '/')[2], split(cMKKeyVaultResourceId, '/')[4])
+//}
+
```
ahmadabdalla, line 267: Option 2 - Inner resource

modules/Microsoft.DBforMySQL/flexibleServers/deploy.bicep
```diff
@@ -0,0 +1,424 @@
...
+  properties: {
+    mode: 'Incremental'
+    template: {
+      '$schema': 'https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#'
+      contentVersion: '1.0.0.0'
+      resources: []
+    }
+  }
+}
+
```
ahmadabdalla, line 262: Option 1 - direct call for the child resource

modules/Microsoft.DBforMySQL/flexibleServers/.test/public/deploy.test.bicep
ahmadabdalla: Can we resolve this @AlexanderSehr ?

modules/Microsoft.DBforMySQL/flexibleServers/deploy.bicep
ahmadabdalla: As discussed. We've moved to the new inner method. I added an issue for aligning other modules as well. See #2838

AlexanderSehr (approved):

PR #2658, opened 2023-02-06
[Utilities] New condition for adding examples in module readme

MariusStorhaug:
# Description

- Updated the condition for when the readme generator adds tests into the readme file as examples. This is now going to run if there are tests (`deploy.test.*`) in the `.test` folder. Current implementation is based on folder levels.

## Pipeline references

| Pipeline |
| - |
| [Batch: BatchAccounts](https://github.com/Azure/ResourceModules/actions/workflows/ms.batch.batchaccounts.yml) |
| [KeyVault: Vaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.keyvault.vaults.yml) |
| [Storage: StorageAccounts](https://github.com/Azure/ResourceModules/actions/workflows/ms.storage.storageaccounts.yml) |

# Type of Change

- [ ] Update to documentation

# Checklist

- [ ] I did format my code


utilities/tools/Set-ModuleReadMe.ps1
```diff
@@ -1566,8 +1566,8 @@ function Set-ModuleReadMe {
         $readMeFileContent = Set-CrossReferencesSection @inputObject
     }
 
-    $isTopLevelModule = $fullModuleIdentifier.Split('/').Count -eq 2 # <provider>/<resourceType>
-    if ($SectionsToRefresh -contains 'Deployment examples' -and $isTopLevelModule) {
+    $hasTests = Get-ChildItem -Path (Join-Path -Path $moduleRoot -ChildPath '.test') -Recurse -Filter 'deploy.test.bicep'
```
AlexanderSehr, line 1569: I'm fairly certain this will break if using ARM or Parameter files. While the later are not too important - ARM is still a valid scenario. So please update the filter accordingly. 
Also, previously it returned a boolean as I wanted to explicitely determine if the handled module was a top-level module, or not. Is there a reason this was changed to a list? 
Last but not least - what's the reason in general? I assume the original implementation did not work for some module?

MariusStorhaug: Good points!
- For 1 and 2. I will update filter to accommodate arm/json + parameter files and maybe add a count -ne 0, to return a boolean.
- For 3. This was for a specific implementation where the folder structure was changed slightly (added one folder level between modules and resource providers). The suggested approach would make it check for the existence of the tests, which seemed to be the deciding factor for when to add examples or not.

MariusStorhaug: Updated to

```
$hasTests = (Get-ChildItem -Path (Join-Path -Path $moduleRoot -ChildPath '.test') -Recurse -Include 'deploy.test.*').count -gt 0
```

modules/Microsoft.DocumentDB/databaseAccounts/.test/test/deploy.test.json
```diff
@@ -0,0 +1,115 @@
+targetScope = 'subscription'
```
AlexanderSehr, line 1: Don't forget to remove

MariusStorhaug: Done

AlexanderSehr (approved):

Issue #2641, opened 2023-02-01
[Bug Report]: Argument of type "null | string" is not assignable to parameter of type "string" warning on many modules


shawntmeyer:
### Describe the bug

Several child extensions report the subject error in VS Code mainly on the name property passed to the resource when using a reference to the parent. Here is an example from the resourceGroups nested_roleAssignments.bicep where the 'resourceId' specified in the name property is generating the error. This is just one example.

```bicep
resource roleAssignment 'Microsoft.Authorization/roleAssignments@2022-04-01' = [for principalId in principalIds: {
  name: guid(last(split(resourceId, '/')), principalId, roleDefinitionIdOrName)
  properties: {
    description: description
    roleDefinitionId: contains(builtInRoleNames, roleDefinitionIdOrName) ? builtInRoleNames[roleDefinitionIdOrName] : roleDefinitionIdOrName
    principalId: principalId
    principalType: !empty(principalType) ? any(principalType) : null
    condition: !empty(condition) ? condition : null
    conditionVersion: !empty(conditionVersion) && !empty(condition) ? conditionVersion : null
    delegatedManagedIdentityResourceId: !empty(delegatedManagedIdentityResourceId) ? delegatedManagedIdentityResourceId : null
  }
}]
```

I am not sure if this is a new linter test in the BICEP extension, but there are two possibilities to mitigate this error.

1. Use the any() function to wrap the value of the referencing value (i.e., 'guid(any(last(split(resourceId, '/'))), principalId, roleDefinitionIdOrName)').

```bicep
name: guid(any(last(split(resourceId, '/'))), principalId, roleDefinitionIdOrName)
```

2. Use a non-null assertion postfix operator (!) just introduced in BICEP (but not documented as of now) that tells the linter that we will never have a null value for this passed variable.

```bicep
name: guid(last(split(resourceId, '/'))!, principalId, roleDefinitionIdOrName)
```
For more information see: [Azure\BICEP Github Issue #9565](https://github.com/Azure/bicep/issues/9565) and 
[Azure\BICEP Github Pull Request #9585](https://github.com/Azure/bicep/pull/9585).

### To reproduce

Open modules\Microsoft.Resources\resourceGroups\.bicep\nested_roleAssignments.bicep in Visual Studio Code and navigate to line 229. The error looks like this:

![CARMLIssue](https://user-images.githubusercontent.com/49066369/216045597-3270a90d-2a70-4da6-8ef3-cdd2d695b30a.jpg)

[...]

PR #2684, opened 2023-02-10
Enabled RBAC role assignment for disk encrpytion set key

AlexanderSehr:
# Description

- Enabled RBAC authorization of the Key used in the Disk Encryption Set. Before it was limited to RBAC.
> NOTE: The solution works with User-Assigned Identities only. The reason is that the System-Assigned identity is not yet available when the role assignment would need to happen

| Pipeline |
| - |
| [Compute: DiskEncryptionSets](https://github.com/Azure/ResourceModules/actions/workflows/ms.compute.diskencryptionsets.yml) |

# Type of Change

- [ ] Update to documentation

### DIFF
--- a/modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_keyVaultPermissions.bicep
+++ b/modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_keyVaultPermissions.bicep
@@ -29,6 +29,10 @@ module userAssignedIdentity 'nested_managedIdentityReference.bicep' = {
   scope: resourceGroup(split(userAssignedIdentityResourceId, '/')[2], split(userAssignedIdentityResourceId, '/')[4])
 }
 
+// =============== //
+// Role Assignment //
+// =============== //
+
 resource keyVaultKeyRBAC 'Microsoft.Authorization/roleAssignments@2022-04-01' = if (rbacAuthorizationEnabled == true) {
   name: guid('msi-${keyVault::key.id}-${location}-${userAssignedIdentityResourceId}-Key-Reader-RoleAssignment')
   scope: keyVault::key
@@ -39,11 +43,17 @@ resource keyVaultKeyRBAC 'Microsoft.Authorization/roleAssignments@2022-04-01' = 
   }
 }
 
-resource accessPolicies 'Microsoft.KeyVault/vaults/accessPolicies@2022-07-01' = if (rbacAuthorizationEnabled != true) {
+// ============= //
+// Access Policy //
+// ============= //
+
+module keyVaultAccessPolicies '../../../Microsoft.KeyVault/vaults/accessPolicies/deploy.bicep' = {
   name: '${uniqueString(deployment().name, location)}-DiskEncrSet-KVAccessPolicies'
-  properties: {
+  params: {
+    keyVaultName: last(split(keyVaultResourceId, '/'))!
     accessPolicies: [
       {
+        tenantId: subscription().tenantId
         objectId: userAssignedIdentity.outputs.principalId
         permissions: {
           keys: [
@@ -52,7 +62,6 @@ resource accessPolicies 'Microsoft.KeyVault/vaults/accessPolicies@2022-07-01' = 
             'unwrapKey'
           ]
         }
-        tenantId: subscription().tenantId
       }
     ]
   }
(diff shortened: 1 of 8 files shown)


modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,68 @@
...
+    roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '12338af0-0e69-4776-bea7-57ae8d297424') // Key Vault Crypto User
+    principalType: 'ServicePrincipal'
+  }
+}
+
+// ============= //
+// Access Policy //
+// ============= //
+
+module keyVaultAccessPolicies '../../../Microsoft.KeyVault/vaults/accessPolicies/deploy.bicep' = {
```
ahmadabdalla, line 50: @AlexanderSehr should this only run when `rbacAuthorizationEnabled` is `false`? As the default is `true` but this means that the access policy will run regardless. Am I understanding this correctly?

modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_roleAssignments.bicep
```diff
@@ -69,7 +69,7 @@ var builtInRoleNames = {
 }
 
 resource diskEncryptionSet 'Microsoft.Compute/diskEncryptionSets@2020-12-01' existing = {
-  name: last(split(resourceId, '/'))
+  name: last(split(resourceId, '/'))!
```
ahmadabdalla, line 72: What is this used for?

modules/Microsoft.Compute/diskEncryptionSets/deploy.bicep
```diff
@@ -60,11 +60,27 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
...
+resource keyVault 'Microsoft.KeyVault/vaults@2021-10-01' existing = {
+  name: last(split(keyVaultResourceId, '/'))!
   scope: resourceGroup(split(keyVaultResourceId, '/')[2], split(keyVaultResourceId, '/')[4])
+
+  resource key 'keys@2021-10-01' existing = {
+    name: keyName
+  }
 }
 
+// Note: This is only enabled for user-assigned identities and the service's system-assigned identity isn't available during its intitial deployment
```
ahmadabdalla, line 72:
```suggestion
// Note: This is only enabled for user-assigned identities as the service's system-assigned identity isn't available during its initial deployment
```

modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_roleAssignments.bicep
AlexanderSehr: Hey @ahmadabdalla this is the new character introduced by the PG to address this warning #2641 . The new Bicep version which is required for this change is rolled out to the GitHub agents now and hence I addressed the warning for my implementation here directly.

modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_keyVaultPermissions.bicep
AlexanderSehr: Good find. I accidently deleted the condition when replacing the plain `resource` deployment with the module. Should be fixed now.
Fun fact: This is how it was implemented originally. We borderline always set the access policies no matter if they are eventually used or not.

modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_roleAssignments.bicep
ahmadabdalla: Thanks for clarifying @AlexanderSehr . Learned something new today 😀

ahmadabdalla (approved):

modules/Microsoft.Compute/diskEncryptionSets/.bicep/nested_roleAssignments.bicep
AlexanderSehr: No problem. It's a nasty nasty world out there :D

PR #2805, opened 2023-02-16
[Modules] Fix incorrect test 'name' parameter

BonnieMcb:
'name' should not be in its own {object}

# Description

> Please include a summary of the change and which issue is fixed.

| Pipeline |
| - |
| |

# Type of Change

- [x] Update to documentation

# Checklist

- [ ] I did format my code

### DIFF
--- a/modules/Microsoft.EventHub/namespaces/readme.md
+++ b/modules/Microsoft.EventHub/namespaces/readme.md
@@ -515,8 +515,6 @@ module namespaces './Microsoft.EventHub/namespaces/deploy.bicep' = {
       "value": [
         {
           "name": "<<namePrefix>>-az-evh-x-001"
-        },
-        {
           "authorizationRules": [
             {
               "name": "RootManageSharedAccessKey",


modules/Microsoft.EventHub/namespaces/readme.md
```diff
@@ -515,8 +515,6 @@ module namespaces './Microsoft.EventHub/namespaces/deploy.bicep' = {
       "value": [
         {
           "name": "<<namePrefix>>-az-evh-x-001"
-        },
```
AlexanderSehr, line 518: Hey @BonnieMcb this is a good catch. But it actually needs to be fixed in the underlying test file: 
https://github.com/Azure/ResourceModules/blob/ac97d3b5d4485792b443ff3c2d12758a103abe0f/modules/Microsoft.EventHub/namespaces/.test/common/deploy.test.bicep#L89-L91
and then we have to re-run the `Set-ModuleReadMe` script to update the ReadMe that you suggested the original fix for.

If you need any support with that, let us know :)

AlexanderSehr: See other comment

AlexanderSehr (commented):
I stand corrected, this is a non-issue. 
The 'EventHubs' is an array parameter with 2 entries. The first only sets the mandatory value (`name`: `'<<namePrefix>>-az-evh-x-001'`), the 2nd sets a bunch of values - also includding a `name` parameter (`'<<namePrefix>>-az-evh-x-002`).

Each entry of the eventhubs array must be of the same type, which in this case is 'object'

AlexanderSehr:
Hey @BonnieMcb please see my comment above. I hope the explanation made sense. If not, feel free to re-open the PR & tag me :)

PR #2813, opened 2023-02-17
[Modules] Removed the paramerers of those properties that allow a single value only & updated docs

AlexanderSehr:
# Description

- Removed the paramerers of those properties that allow a single value only
- Updated docs
- Removed 'minPrem' test from storage account to unblock tests (and because the feature isn't available in our test location West Europe)

| Pipeline |
| - |
| [KeyVault - Vaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.keyvault.vaults.yml) |
| [Resources - Tags](https://github.com/Azure/ResourceModules/actions/workflows/ms.resources.tags.yml) |
| [ServiceBus - Namespaces](https://github.com/Azure/ResourceModules/actions/workflows/ms.servicebus.namespaces.yml) |
| [Sql - ManagedInstances](https://github.com/Azure/ResourceModules/actions/workflows/ms.sql.managedinstances.yml) |
| [Storage - StorageAccounts](https://github.com/Azure/ResourceModules/actions/workflows/ms.storage.storageaccounts.yml) |

# Type of Change

- [ ] Update to documentation


modules/Microsoft.Storage/storageAccounts/blobServices/deploy.bicep
```diff
@@ -57,7 +54,7 @@ param diagnosticMetricsToEnable array = [
 ]
 
 @description('Optional. The name of the diagnostic setting, if deployed.')
-param diagnosticSettingsName string = '${name}-diagnosticSettings'
+param diagnosticSettingsName string = 'default-diagnosticSettings'
```
eriqua, line 57: I'd keep the optional diagnostic settings feature as aligned as possible across all modules. In this and other cases below I'd suggest using a variable `name` since the hardcoded value is used in more than one place

AlexanderSehr: hm ok

AlexanderSehr: Just tried and actually - no. We cannot reference variables in parameters.

eriqua: Just opened a draft PR #2814 with what I meant

AlexanderSehr: Looks good. The only 'caviat' is, that now the user is actually not aware of this default value as it is not published to the readme.

eriqua: True, in that case we should probably add something in the description, e.g., "defaults to ... if empty"? In case this alternative makes sense, I guess it should be implemented in a consistent way also for modules not having an hardcoded name? Maybe worth a quick internal discussion.

AlexanderSehr: Following a discussion we'll go with the suggestion by @eriqua to keep the diagnostic settings aligned & reduce the amount of duplicated code/values

AlexanderSehr: We can resolve this comment once the PR was merged into this

modules/Microsoft.Storage/storageAccounts/queueServices/readme.md
```diff
@@ -35,11 +35,10 @@ This module can be used to deploy a file share service into a storage account.
 | `diagnosticLogCategoriesToEnable` | array | `[allLogs]` | `[allLogs, StorageDelete, StorageRead, StorageWrite]` | The name of logs that will be streamed. "allLogs" includes all possible logs for the resource. |
 | `diagnosticLogsRetentionInDays` | int | `365` |  | Specifies the number of days that logs will be kept for; a value of 0 will retain data indefinitely. |
 | `diagnosticMetricsToEnable` | array | `[Transaction]` | `[Transaction]` | The name of metrics that will be streamed. |
-| `diagnosticSettingsName` | string | `[format('{0}-diagnosticSettings', parameters('name'))]` |  | The name of the diagnostic setting, if deployed. |
+| `diagnosticSettingsName` | string | `'default-diagnosticSettings'` |  | The name of the diagnostic setting, if deployed. |
```
eriqua, line 38: I guess this should be updated as well?

AlexanderSehr: good point. Done

AlexanderSehr: This will become real interesting for the Rest2Carml automation :D But I think by now with some of the constructs we have (like the necessary use of union) we cannot use it to update much anymore anyways :o

docs/wiki/The library - Module design.md
```diff
@@ -463,7 +465,7 @@ Within a bicep file, use the following conventions:
 
 ## Modules
 
-- Module symbolic names are in camel_Snake_Case, following the schema `<mainResourceType>_<referencedResourceType>` e.g., `storageAccount_fileServices`, `virtualMachine_nic`, `resourceGroup_roleAssignments`.
+- Module symbolic names are in camel*Snake_Case, following the schema `<mainResourceType>*<referencedResourceType>`e.g.,`storageAccount_fileServices`, `virtualMachine_nic`, `resourceGroup_roleAssignments`.
```
eriqua, line 468: I'm not sure I fully grasp the expression here. Examples make it clear, but what does the * stand for? Shouldn't the schema be `<mainResourceType>_<referencedResourceType>`?

docs/wiki/The library - Module design.md
```diff
@@ -391,6 +391,8 @@ Within a bicep file, use the following conventions:
   - `Conditional` - The parameter value can be optional or required based on a condition, mostly based on the value provided to other parameters.
   - `Optional` - The parameter value is not mandatory. The module provides a default value for the parameter.
   - `Generated` - The parameter value is generated within the module and should not be specified as input.
+- If parameters map to resource properties, they should not contain the resource's name as it would introduce redundancy. For example, the `name` property of the Key Vault module should be just that and not `keyVaultName`. The rationale is that the consumers know that the name is for the Key Vault if they deploy its module.
+- Further, if a property value allows a single value only, there is no need to introduce a parameter for it. Instead it can be hardcoded into the deployment. For example, the name of a blobServices resource can only be `default`. Hence we can implement its name property directly as `name: 'default'`.
```
eriqua, line 395: Far from easy, but should we update the paragraph to include the updated approach i.e., using the variable name and make the diagnosticName empty?

eriqua: I'm also thinking that using a variable for the hardcoded value even when diagnostic is not supported would be easier to describe and provide guidelines for. I sense you won't agree, passing you the ball😜

docs/wiki/The library - Module design.md
AlexanderSehr: This was some damn extension. And I still did not find out which. This must be undone

docs/wiki/The library - Module design.md
AlexanderSehr: I definitely wouldn't agree doing it 😉 but yeah, describing it would be easier 😄
All add a "special case" entry here. I hope thought that we won't end up with too many in the long term. Already many contributions seem to ignore the existing ones

AlexanderSehr: Done

AlexanderSehr: For 'all' it will anyways become quite interesting as I believe for some resources we already apply some logic to the name which we now need to intertwine with this.

AlexanderSehr: So I'm not sure how to update the description for that. Or do you already have something in mind?
I can also update the other diagnostic settings in this PR if you want :)

eriqua: Let's keep them separated as this PR is done already (will just have a final review). How to update the wiki will become more clear once we update others. This is correct for the current implementation so I think is good

eriqua (approved):
🆗

Issue #2613, opened 2023-01-24
[Bug Report]: Issues deploying Synapse with required Storage Parameters


chris-gibbs:
### Describe the bug

Getting a validation error when deploying Synapse when providing only the defaultDataLakeStorageAccountName and defaultDataLakeStorageFilesystem parameters as per the examples.

Tested with commit from 13/01/2023. Not a specific release though.

### To reproduce

Deploy synapse with managed vnet and CMK for storage.

only provide 
- defaultDataLakeStorageAccountName 
- defaultDataLakeStorageFilesystem

Fix
Update file https://github.com/Azure/ResourceModules/blob/main/modules/Microsoft.Synapse/workspaces/deploy.bicep

Add a new param 'resourceId' under 'defaultDataLakeStorage' for resource workspace 'Microsoft.Synapse/workspaces@2021-06-01'

Fixed module call

```bicep
module synapse '../../../../carml/modules/Microsoft.Synapse/workspaces/deploy.bicep' = {
  name: 'deploy_synapse_${env}'
  scope: rgSynapse
  params: {
    name: synapseWorkspaceName
    location: location
    **defaultDataLakeStorageID: storage.outputs.resourceId**
    defaultDataLakeStorageAccountName: storageAccountName
    defaultDataLakeStorageFilesystem:defaultDataLakeStorageFilesystem
    defaultDataLakeStorageCreateManagedPrivateEndpoint: true
    azureADOnlyAuthentication: true
    initialWorkspaceAdminObjectID: adminGroupID
    managedResourceGroupName: '${rgSynapseName}-managed'
    managedVirtualNetwork: true
    allowedAadTenantIdsForLinking: [subscription().tenantId]
    preventDataExfiltration: true
    publicNetworkAccess:'Disabled'
    // encryptionActivateWorkspace: true // todo: raise bug. This doesnt work when the KV instance is in a different RG. 
    encryption: true
    cMKUseSystemAssignedIdentity: true
    cMKKeyVaultResourceId: kv.id
    cMKKeyName: cMKKeyName
    sqlAdministratorLogin: sqlAdministratorLogin
    sqlAdministratorLoginPassword: sqlAdministratorLoginPassword
    privateEndpoints: [
      {
        name: '${synapseWorkspaceName}-dev-pe'
        service: 'dev'
        subnetResourceId: privateEndpointSubnetId
        privateDnsZoneGroup: {
          privateDNSResourceIds: [
[...]

PR #2825, opened 2023-02-20
[Modules] Synapse workspaces Added support for `resourceId` #2613

Lddeiva:
# Description

> Please include a summary of the change and which issue is fixed.
- The field `resourceId` was added in the API which wasn't updated in the module. When the module is used, it is showing errors as it couldn't find the `resourceId` which is expected by the API.

| Pipeline |
| - |
| |

# Type of Change

- [x] Update to documentation

# Checklist

- [x] I did format my code

### DIFF
--- a/modules/Microsoft.Synapse/workspaces/deploy.bicep
+++ b/modules/Microsoft.Synapse/workspaces/deploy.bicep
@@ -197,7 +197,7 @@ resource workspace 'Microsoft.Synapse/workspaces@2021-06-01' = {
     } : null
     defaultDataLakeStorage: {
       resourceId: defaultDataLakeStorageAccountResourceId
-      accountUrl: 'https://${split(defaultDataLakeStorageAccountResourceId, '/')[-1]}.dfs.${environment().suffixes.storage}'
+      accountUrl: 'https://${last(split(defaultDataLakeStorageAccountResourceId, '/'))!}.dfs.${environment().suffixes.storage}'
       filesystem: defaultDataLakeStorageFilesystem
       createManagedPrivateEndpoint: managedVirtualNetwork ? defaultDataLakeStorageCreateManagedPrivateEndpoint : null
     }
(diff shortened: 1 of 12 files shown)


Lddeiva:
Fixes #2613

modules/Microsoft.Synapse/workspaces/deploy.bicep
```diff
@@ -15,6 +15,9 @@ param azureADOnlyAuthentication bool = false
 @description('Optional. AAD object ID of initial workspace admin.')
 param initialWorkspaceAdminObjectID string = ''
 
+@description('Required. Resource ID of the default ADLS Gen2 storage account.')
+param defaultDataLakeStorageAccountResourceId string
```
eriqua, line 19: Is there any constraint to where the storage should be placed? Is it required to be in the same subscription and resource group as the main resource?
If so, we could derive the resource id from the input name as an existing resource. Otherwise, 
I guess we should get rid of the below storage account name and derive it from the new resource ID input parameter. What's your take on that?

Lddeiva: Thanks for the suggestion @eriqua . I removed the `defaultDataLakeStorageAccountName` and calculating the name using `defaultDataLakeStorageAccountResourceId`.

eriqua (commented):
Thanks for your contribution @Lddeiva, looks very good. Do you happen to have a pipeline badge reference to attach to this PR to prove its successful run?

Lddeiva:
@eriqua , I'm unable to queue the workflow runs. I assume only the maintainers can queue them.

image

Lddeiva:
@eriqua, I have issues in setting up the CI environment. Can you please help me set them up?

eriqua:
> @eriqua, I have issues in setting up the CI environment. Can you please help me set them up?

Sure @Lddeiva let's address offline sometime next week

PR #2831, opened 2023-02-21
[Modules] Updated  Microsoft.Compute/disks: API versions, image sku, new properties

### LINKED ISSUE #2827: [Failed pipeline] Compute - Disks
failed run: https://github.com/Azure/ResourceModules/actions/runs/4226155599

krbar:
# Description

This PR resolves #2827 and introduces following improvements:

- replaced a non-existing source image sku for the "Import" test case: 
  - old: `Windows-10` - `19h2-evd`
  - new: `Windows-11` - `win11-21h2-avd`
- updated API versions of the test deployments
- updated API versions of the `disk` resource and related improvements
  - added new disk sku: `PremiumV2_LRS`
  - added new, non-breaking parameters:
    - `architecture`
    - `optimizedForFrequentAttach`

> Note: two additional parameters included in the api version [2022-07-02](https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/2022-07-02/disks?pivots=deployment-language-bicep) were not included in the module update as they appear to be in preview and/or are not supported in the test subscription and could not be tested: `disk.supportedCapabilities.diskControllerTypes` and `PerformancePlus`

| Pipeline |
| - |
| [Compute - Disks](https://github.com/Azure/ResourceModules/actions/workflows/ms.compute.disks.yml) |

# Type of Change

# Checklist

- [x] I did format my code

### DIFF
--- a/modules/Microsoft.Compute/disks/.test/image/deploy.test.bicep
+++ b/modules/Microsoft.Compute/disks/.test/image/deploy.test.bicep
@@ -23,7 +23,7 @@ param enableDefaultTelemetry bool = true
 
 // General resources
 // =================
-resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
+resource resourceGroup 'Microsoft.Resources/resourceGroups@2022-09-01' = {
   name: resourceGroupName
   location: location
 }
--- a/modules/Microsoft.Compute/disks/.bicep/nested_roleAssignments.bicep
+++ b/modules/Microsoft.Compute/disks/.bicep/nested_roleAssignments.bicep
@@ -68,7 +68,7 @@ var builtInRoleNames = {
   'Windows Admin Center Administrator Login': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'a6333a3e-0164-44c3-b281-7a577aff287f')
 }
 
-resource disk 'Microsoft.Compute/disks@2021-08-01' existing = {
+resource disk 'Microsoft.Compute/disks@2022-07-02' existing = {
   name: last(split(resourceId, '/'))!
 }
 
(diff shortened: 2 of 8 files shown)


modules/Microsoft.Compute/disks/.test/image/deploy.test.bicep
```diff
@@ -23,7 +23,7 @@ param enableDefaultTelemetry bool = true
 
 // General resources
 // =================
-resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
+resource resourceGroup 'Microsoft.Resources/resourceGroups@2022-09-01' = {
```
eriqua, line 26: Thanks for pointing this out and it definitely makes sense to me. I'm just wondering if it wouldn't be better to park API version update on dependency resources for now and manage that in bulk for all modules in a separated PR to avoid inconsistencies across the library. At least for the most used resources such as rg, storage, kv and vnet. What's your take on that?

krbar: I think that it may be not be easy to keep the consistency across the library that way, for at least two reasons:
- I just searched in the library and there are already different API versions there for the popular resource types mentioned above
- while it's relatively easy to avoid changing API versions in existing modules, creation of new ones would require the contributors to explicitly use older API versions, which they probably don't even know

I agree that we should have a separate PR updating API versions for (at least) popular dependencies, but I think that (thank to the powerful Search/Replace function of VS Code) the complexity of this task will not significantly increase when we must search for different existing API versions. 

I'm happy to follow your recommendation, it's not a big deal to revert these updates.

krbar: As agreed in the sync call, resolving the conversation.

eriqua (approved):
🆗

PR #2893, opened 2023-03-10
[Fixes] Storage Account Local User : Pass on correct parameter

davidkarlsen:
# Description

The parameter is not passed on as expected, but seems to be a copy/paste error.
In effect I cannot enable the authorizedKeys for the local user.

need to come back after it has run

| Pipeline |
| - |
| |

# Type of Change

# Checklist

- [X] I did format my code

### DIFF
--- a/modules/Microsoft.Storage/storageAccounts/deploy.bicep
+++ b/modules/Microsoft.Storage/storageAccounts/deploy.bicep
@@ -378,7 +378,7 @@ module storageAccount_localUsers 'localUsers/deploy.bicep' = [for (localUser, in
     storageAccountName: storageAccount.name
     name: localUser.name
     hasSharedKey: contains(localUser, 'hasSharedKey') ? localUser.hasSharedKey : false
-    hasSshKey: contains(localUser, 'hasSshPassword') ? localUser.hasSshPassword : true
+    hasSshKey: contains(localUser, 'hasSshKey') ? localUser.hasSshKey : false
     hasSshPassword: contains(localUser, 'hasSshPassword') ? localUser.hasSshPassword : false
     homeDirectory: contains(localUser, 'homeDirectory') ? localUser.homeDirectory : ''
     permissionScopes: contains(localUser, 'permissionScopes') ? localUser.permissionScopes : []
--- a/modules/Microsoft.Storage/storageAccounts/deploy.bicep
+++ b/modules/Microsoft.Storage/storageAccounts/deploy.bicep
@@ -377,11 +377,11 @@ module storageAccount_localUsers 'localUsers/deploy.bicep' = [for (localUser, in
   params: {
     storageAccountName: storageAccount.name
     name: localUser.name
+    hasSshKey: localUser.hasSshKey
+    hasSshPassword: localUser.hasSshPassword
+    permissionScopes: localUser.permissionScopes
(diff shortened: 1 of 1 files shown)


modules/Microsoft.Storage/storageAccounts/deploy.bicep
```diff
@@ -378,7 +378,7 @@ module storageAccount_localUsers 'localUsers/deploy.bicep' = [for (localUser, in
     storageAccountName: storageAccount.name
     name: localUser.name
     hasSharedKey: contains(localUser, 'hasSharedKey') ? localUser.hasSharedKey : false
-    hasSshKey: contains(localUser, 'hasSshPassword') ? localUser.hasSshPassword : true
+    hasSshKey: contains(localUser, 'hasSshKey') ? localUser.hasSshKey : false
```
AlexanderSehr, line 381: Good catch. It is however actually a 'required' value so it should be passed as follows:
```suggestion
    hasSshKey: localUser.hasSshKey
```

davidkarlsen: seems odd to only do that for the one parameter and not all of them, but ok.

AlexanderSehr: Hey David,
that is true. I only commented on the line that was part of your PR. For the child-module deployment in general, all required parameter should be passed down as such:
- hasSshKey
- hasSshPassword
- name
- permissionScopes

Which means, the entire block should look like:
```bicep
module storageAccount_localUsers 'localUsers/deploy.bicep' = [for (localUser, index) in localUsers: {
  name: '${uniqueString(deployment().name, location)}-Storage-LocalUsers-${index}'
  params: {
    storageAccountName: storageAccount.name
    name: localUser.name
    hasSshKey: localUser.hasSshKey
    hasSshPassword: localUser.hasSshPassword
    permissionScopes: localUser.permissionScopes
    hasSharedKey: contains(localUser, 'hasSharedKey') ? localUser.hasSharedKey : false
    homeDirectory: contains(localUser, 'homeDirectory') ? localUser.homeDirectory : ''
    sshAuthorizedKeys: contains(localUser, 'sshAuthorizedKeys') ? localUser.sshAuthorizedKeys : []
    enableDefaultTelemetry: enableReferencedModulesTelemetry
  }
}]
```

However, why for example `hasSharedKey` isn't treated the same way, I don't know. The original contributor only considered the initial list as 'mandatory' values - which doesn't mean we could not change it.

davidkarlsen: Added, but we're maybe a bit out of scope of the original bugfix now... Can we get the CI to run and merge this now?

AlexanderSehr: Hey @davidkarlsen as per the [Contribution Guide](https://github.com/Azure/ResourceModules/wiki/Contribution%20guide%20-%20Contribution%20flow) we would usually ask the contributor to setup the CI in their fork and provide a pipeline badge that shows that the contribution works. Unless folks don't have access to a subscription, that is.
This may change in the future and we may be able to enable users to run pipelines directly. But for the time being, we follow the flow description in the contribution guide.

That being said, the change is so small that it appears highly unlikely that it will fail. So we can skip that part.

AlexanderSehr (approved):

Issue #2996, opened 2023-03-15
[Feature Request]: StorageAccounts - Allow Resource level policy exemption


shawntmeyer:
### Description

The current policyExemption modules only support exemptions at the MG, Sub, or RG levels. When policyExemptions are deployed to resources, they are treated as extensions. We need this capability to allow the creation of a storage account with public access when we have a policy applied to the Sub or MG that blocks this access.

rahalan:
Team will revisit the issue next week

AlexanderSehr:
Blocking the overall implementation for the moment until we were able to sort out the remaining questions in the PR. If the PR is rejected afterall, this issue will be closed too - if not (i.e., it is merged), we may move it out of blocked again.

PR #2997, opened 2023-03-15
[Module] Microsoft.Storage/StorageAccounts policy exemption

shawntmeyer:
# Description

This update adds the deployment of policy exemptions at the Storage Account level which is a useful deployment orchestration when we must deploy storage accounts with public access allowed and we have policy deployed at a higher scope that blocks that resource creation.

## Pipeline references

[Storage - StorageAccounts](https://github.com/shawntmeyer/CARML/actions/workflows/ms.storage.storageaccounts.yml)

# Type of Change

- [X] Update to documentation

# Checklist

- [X] I did format my code

### DIFF
--- a/modules/Microsoft.Storage/storageAccounts/.bicep/nested_policyExemptions.bicep
+++ b/modules/Microsoft.Storage/storageAccounts/.bicep/nested_policyExemptions.bicep
@@ -1,0 +1,63 @@
+@sys.description('Required. The resource ID of the resource to apply the policy exemption to.')
+param resourceId string
+
+@sys.description('Required. Specifies the name of the policy exemption. Maximum length is 64 characters.')
+@maxLength(64)
+param name string
+
+@sys.description('Optional. The display name of the policy exemption. Maximum length is 128 characters.')
+@maxLength(128)
+param displayName string = ''
+
+@sys.description('Optional. The description of the policy exemption.')
+param description string = ''
+
+@sys.description('Optional. The policy exemption metadata. Metadata is an open ended object and is typically a collection of key-value pairs.')
+param metadata object = {}
+
+@sys.description('Optional. The policy exemption category. Possible values are Waiver and Mitigated. Default is Mitigated.')
+@allowed([
+  'Mitigated'
+  'Waiver'
+])
+param exemptionCategory string = 'Mitigated'
+
+@sys.description('Required. The resource ID of the policy assignment that is being exempted.')
+param policyAssignmentId string
+
+@sys.description('Optional. The policy definition reference ID list when the associated policy assignment is an assignment of a policy set definition.')
+param policyDefinitionReferenceIds array = []
+
+@sys.description('Optional. The expiration date and time (in UTC ISO 8601 format yyyy-MM-ddTHH:mm:ssZ) of the policy exemption. e.g. 2021-10-02T03:57:00.000Z.')
+param expiresOn string = ''
+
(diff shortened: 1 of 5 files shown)


ahmadabdalla:
@shawntmeyer This is a new feature that we are introducing like how we do RBAC on resource level.

may I suggest we put this one on hold to bring in this topic into one of internal discussions. I will tag the issue https://github.com/Azure/ResourceModules/issues/2996. as well. 
CC: @rahalan

ahmadabdalla (commented):
Placeholder comment until feature is discussed

AlexanderSehr (commented):
Create stuff in any case. The only caviats I'm seeing are
- Increased file size (paying into the 4 MB limit)
- In theory there could be one excemption per resource type, I guess

Also I hope normal contributors won't have the permissions to just deploy excemptions 😄

eriqua:
Marked as draft as per offline discussion

modules/Microsoft.Storage/storageAccounts/deploy.bicep
```diff
@@ -338,6 +349,21 @@ module storageAccount_roleAssignments '.bicep/nested_roleAssignments.bicep' = [f
   }
 }]
 
+resource storageAccount_policyExemptions 'Microsoft.Authorization/policyExemptions@2022-07-01-preview' = [for policyExemption in policyExemptions: if (!empty(policyExemptions)) {
```
AlexanderSehr, line 352: Wouldn't the storage account deployment be already blocked before we even make it to the excemption?

shawntmeyer: it is not, the storage account is deployed. The common tests for both storage accounts and key vaults deploy a policy assignment that deny's creation of a resource that doesn't meet the policy.  I verified that the test is a true test in that the resource doesn't meet the policy, but the exception works and the deployment succeeds.

modules/Microsoft.Storage/storageAccounts/.bicep/nested_policyExemptions.bicep
```diff
@@ -0,0 +1,63 @@
+@sys.description('Required. The resource ID of the resource to apply the policy exemption to.')
```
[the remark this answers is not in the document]
shawntmeyer: removed file

modules/Microsoft.Storage/storageAccounts/deploy.bicep
eriqua: Is the policy a deny or an audit policy?

shawntmeyer: @eriqua, it is a Deny policy (https://portal.azure.com/#blade/Microsoft_Azure_Policy/PolicyDetailBlade/definitionId/%2Fproviders%2FMicrosoft.Authorization%2FpolicyDefinitions%2Fc9d007d0-c057-4772-b18c-01e546713bcd) and in the test deployed with the Deny effect. Same for Key Vault.

shawntmeyer: This is exactly why I needed the PR. We can not create Resource Level exemptions without the resource ID and we can't create the resource with a deny policy in place without the exemption. The only way to do this is through ARM as a single deployment. ARM sorts it out and allows the resource to be created due to the exemption. Without this we would have to temporarily place an exemption at the RG level or disable the policy.

AlexanderSehr:
Marking as draft until we found the best way to integrate this

PR #2999, opened 2023-03-15
[Fixes] Removed child resource that may not be used in the applied context

AlexanderSehr:
# Description

The configuration child may not be used this way as documented e.g. [here](https://learn.microsoft.com/en-us/azure/templates/microsoft.dbformysql/flexibleservers/configurations?pivots=deployment-language-bicep).

| Pipeline |
| - |
| |

# Type of Change

- [ ] Update to documentation


eriqua (approved):
🆗

PR #3260, opened 2023-05-17
[Modules] Storage Accounts module - added 'accessTier' parameter for file shares

### LINKED ISSUE #3256: [Feature Request from Discussions] Azure file Share AccessTier
### Discussed in https://github.com/Azure/ResourceModules/discussions/3188

<div type='discussions-op-text'>

<sup>Originally posted by **dbeaupre** May  9, 2023</sup>
Hi, 

It's me or it's not possible to set the AccessTier of a fileshare using the actual storage account module version ? I dont see any accessTier property... Thanks</div>

CC: @dbeaupre

krbar:
# Description

PR resolves #3256 and adds a new parameter `accessTier` to the file shares module.

## Pipeline references

| Pipeline |
| - |
| [Storage - StorageAccounts](https://github.com/Azure/ResourceModules/actions/workflows/ms.storage.storageaccounts.yml) |

# Type of Change

# Checklist

- [x] I did format my code


modules/Storage/storageAccounts/fileServices/main.bicep
```diff
@@ -135,6 +137,7 @@ module fileServices_shares 'shares/main.bicep' = [for (share, index) in shares:
     storageAccountName: storageAccount.name
     fileServicesName: fileServices.name
     name: share.name
+    accessTier: contains(share, 'accessTier') ? share.accessTier : defaultShareAccessTier
```
eriqua, line 140: Did you by any chance test what happens if `null` is passed here instead of the `defaultShareAccessTier` variable? I wonder if that would cover the default behavior already

AlexanderSehr: @krbar

krbar: @eriqua 
It's a tricky one, because `accessTier` is mandatory in case of `FileStorage` storage account and optional in case of general purpose storage account. So I've created a conditional parameter for it, which defaults to `TransactionOptimized`, which is fine for general purpose storage and invalid for `FileStorage`.

This means that setting a `null` here will in fact not use `null` in the deployment, but `TransactionOptimized` (default of the module parameter). This results in an error if the user doesn't set the `accessTier` for the `FileStorage` storage accounts. The problem is that the error message will be quite meaningless: "The value for one of the HTTP headers is not in the correct format".

So I think it is better to keep the `defaultShareAccessTier` logic, unless we change the design of the `accessTier` parameter. I will ping you to discuss it.

krbar: @eriqua As discussed today, after some additional (and failed) tests with an empty default value of the `accessTier` parameter, I suggest to keep the current implementation.

eriqua (approved):
🆗

PR #3409, opened 2023-07-10
[Modules] Updated ServiceBus Namespace  NetworkRuleSets Module - Fixed bug in variable object 

lsnoddy:
# Description

Fixed bug in networkRules var object subnet id reference and updated the test cases and readme files for this change. This fixes the broken pipeline and aligns the networkRuleSets module with the documentation (https://learn.microsoft.com/en-us/azure/templates/microsoft.servicebus/namespaces/networkrulesets?pivots=deployment-language-bicep)

| Pipeline |
| - |
[ServiceBus - Namespaces](https://github.com/lsnoddy/ResourceModules/actions/workflows/ms.servicebus.namespaces.yml)
| |

# Type of Change

- [ x ] Bugfix (non-breaking change which fixes an issue)

# Checklist

- [ x ] I'm sure there are no other open Pull Requests for the same update/change
- [ x ] My corresponding pipelines / checks run clean and green without any errors or warnings
- [ x ] My code follows the style guidelines of this project
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding changes to the documentation (readme)
- [ x ] I did format my code


modules/service-bus/namespaces/.test/common/main.test.bicep
```diff
@@ -84,8 +84,8 @@ module testDeployment '../../main.bicep' = {
       trustedServiceAccessEnabled: true
       virtualNetworkRules: [
         {
+          ignoreMissingVnetServiceEndpoint: true
           subnet: {
-            ignoreMissingVnetServiceEndpoint: true
```
AlexanderSehr, line 88: Hey @lsnoddy, I believe the fix has to be the other way around. The property should be "subnet ResourceId" instead of an object with just one property ID.

AlexanderSehr: As such, the template can stay the same 😊

AlexanderSehr: It's an approach we implemented usually in cases where an object only has one property

AlexanderSehr (approved):

PR #3413, opened 2023-07-12
[Modules] Simplified DNS Forwarding Rule Input Parameter

AlexanderSehr:
# Description

- Simplified DNS Forwarding Rule Input Parameter  
  - The parameter object had only one field 'id' - so in alignement with other modules we expect from the user just an array of ids instead of objects and create the object for them

| Pipeline |
| - |
| [Network - DNS Forwarding Rulesets](https://github.com/Azure/ResourceModules/actions/workflows/ms.network.dnsforwardingrulesets.yml) |

# Type of Change

- [ ] Update to documentation


eriqua (approved):
🆗

PR #3579, opened 2023-08-15
[Fixes] ADO pipeline deployment stage condition

tyconsulting:
# Description

Fixing the condition for the deployment stage of the ADO pipeline. the deployment stage should be started when the validation stage status is succeeded. having the condition for "not equals to failed" means users can manually cancel the validation stage to bypass validation. When the validation is manually cancelled, all subsequent stages should be skipped.

I can't post a test run status here because it's not a GitHub Action change. I have tested this change with in a customer's environment with @ahmadabdalla 

# Type of Change

- [ ] Update to documentation

# Checklist

- [x] I did format my code


.azuredevops/pipelineTemplates/stages.module.yml
```diff
@@ -15,7 +15,7 @@ stages:
 
   - stage: deployment
     displayName: Deployment validation
-    condition: and(eq('${{ parameters.deploymentValidation }}', 'True'), ne(dependencies.validation.result, 'Failed'))
+    condition: and(eq('${{ parameters.deploymentValidation }}', 'True'), succeeded(validation))
```
AlexanderSehr, line 18: Hey @tyconsulting, thank you for your PR. A few thoughts from my side:

Our goal was the following state:
- Our main concern when it comes to validation is the Publishing. I.e. all dependencies MUST have run and worked in order to publish (hence the `succeeded()` there)
- Regarding the 'deployment validation' we usually only want to run if the static validation was successfull - Unless - somebody explicitely declared to only run the deployment validation when triggering the pipeline.

We the proposed change - would it still be possible to only run the deployment validation and skip the static validation? In other words: Does `successed(validation)` accepted a stage outcome / result of 'Skipped'?

I believe that's why we checked for 'not failed', as this included both 'succeeded' & 'skipped'. 

Looking forward to your response :)

tyconsulting: hi @AlexanderSehr sorry I may have accidentally deleted my previous response...
by having `succeeded(validation)` as a condition, the static validation must be completed successful for the deployment validation to start. Why would customers only want to run deployment validation without the previous step? since the previous step involves bicep build, it would pick up any bicep syntax errors if the template fails to compile, I'd think it should always be completed first? with proposed change, If the static validation stage is skipped, then deployment validation step will not run.

What I have experienced in customer's environment is that with the current construct, if someone manually clicked on the cancel button for the static validation stage, the pipeline will continue to the deployment validation stage, then to the publish stage (since the publish stage condition is `succeeded()` which means the previous stage must be successful. This gives users a way to bypass static validation and have the modules published without it. It should not be the case. Modules should not be published without passing all the validations.

AlexanderSehr: Hey @tryconsulting, thanks for the reply.

I 100% agree, it should NOT be possible to skip any validation when it comes to publishing. If that's the case, it is an unexapected and undesired behavior that must be fixed. I will try to reproduce it on our end and experiment to understand the behavior batter. Maybe it could also be solved by adding an additional dependency (which should then ideally be picked up by the `succeeded()` too and now allow 'skipped'. 

That being said - the reasons we updated the pipeline to allow a user to independently run either the static or deployment validation are
- That the static validation (including PSRule) can sometimes take longer for bigger modules
- Require a constant update of the readme & main.json which can become quite cumbersome if you just want to test something
- Can (in rare cases [just head one today, hence I'm mentioning it]) block the deployment tests because of some deployment-unrelated issue.

None of these reasons should allow a user to bypass anything. In the past it just happened that each time we wanted to test just specific parts, one would need to manually update the pipeline file to skip parts. So as such it's just a convenient tool.

So again, I agree with your statement and will look into it. But please don't ask me to take the possiblity away to only test the deployment of a template :D

tyconsulting: image
This is what I can share, as you can see I had to cancel both stages manually.

AlexanderSehr: Hey @tyconsulting, I created and tested the following change #3583 . Would that work for you as well?

tyconsulting: hi @AlexanderSehr yeah that would also work. thanks

PR #3923, opened 2023-09-06
[Modules] Updated Data-Protection Vault API version

AlexanderSehr:
# Description

- Updated Data-Protection Vault API version

### Pipeline references

| Pipeline |
| - |
| [DataProtection - BackupVaults](https://github.com/Azure/ResourceModules/actions/workflows/ms.dataprotection.backupvaults.yml) |

## Type of Change

- [ ] Update to documentation

### DIFF
--- a/modules/data-protection/backup-vault/main.bicep
+++ b/modules/data-protection/backup-vault/main.bicep
@@ -54,6 +54,12 @@ param azureMonitorAlertSettingsAlertsForAllJobFailures string = 'Enabled'
 @description('Optional. List of all backup policies.')
 param backupPolicies array = []
 
+@description('Optional. Security settings for the backup vault.')
+param securitySettings object = {}
+
+@description('Optional. Feature settings for the backup vault.')
+param featureSettings object = {}
+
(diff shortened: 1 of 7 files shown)


modules/data-protection/backup-vault/main.bicep
```diff
@@ -54,6 +54,12 @@ param azureMonitorAlertSettingsAlertsForAllJobFailures string = 'Enabled'
 @description('Optional. List of all backup policies.')
 param backupPolicies array = []
 
+@description('Optional. Security settings for the backup vault.')
+param securitySettings object = {}
+
+@description('Optional. Feature settings for the backup vault.')
+param featureSettings object = {}
```
ahmadabdalla, line 61: I know the doco is outdated, but if we are introducing two new params, is there something we can add them to the test file at least so it can be visible the readme on how to use them?

AlexanderSehr: Hey @ahmadabdalla , I wish, that was plan A. But I didn't find a single example (not even in the REST documention, which is usually a great reference).

ahmadabdalla (approved):

PR #3925, opened 2023-09-06
Updated Purview target location

AlexanderSehr:
# Description

- Updated Purview region to US as some defaults (e.g. WestEurope) are not supported

### Pipeline references

| Pipeline |
| - |
| [Purview - Accounts](https://github.com/Azure/ResourceModules/actions/workflows/ms.purview.accounts.yml) |

## Type of Change

- [ ] Update to documentation


ahmadabdalla:
What happened here :S :S How was this working in the past

AlexanderSehr:
> What happened here :S :S How was this working in the past

Wasn't able to find out. It has been like this for a while now and I wouldn't know either. But in the docs it explicitely calls the mentioned locations out. No idea :/

ahmadabdalla (approved):

PR #3984, opened 2023-09-16
[Modules] Added module for Redis Cache Enterprise and Database

ahmadabdalla:
# Description

Created a new module for Redis Cache Enterprise and Database child resource.
Added 3 test cases to validate parameter combinations:
- Common: Deploys a Cluster with Private endpoints and cache persistence
- Min: Minimum requirements.
- Geo Replication: To support an enterprise cluster with Geo replication scenarios.

## Common

image

## Min

image

## Geo replication

image

### Pipeline references

image
image
image
image

| Pipeline |
| - |
| New released Pipeline, tested via Azure DevOps |

## Type of Change

## Checklist

- [ ] I did format my code


modules/cache/redis-enterprise/.test/common/main.test.bicep
```diff
@@ -0,0 +1,124 @@
...
+        }
+      }
+    ]
+    databases: [
+      {
+        clusteringPolicy: 'EnterpriseCluster'
+        evictionPolicy: 'AllKeysLFU'
+        modules: [
+          {
+            name: 'RedisBloom'
```
AlexanderSehr, line 107: Would you think it makes sense to simplify this format to a flat array and format the object for the user? Only makes sense to me if we don't anticipate additional properties...

modules/cache/redis-enterprise/main.bicep
```diff
@@ -0,0 +1,232 @@
...
+
+@description('Optional. The name of logs that will be streamed. "allLogs" includes all possible logs for the resource. Set to \'\' to disable log collection.')
+@allowed([
+  ''
+  'allLogs'
+  'ConnectionEvents'
+  'audit'
+])
+param diagnosticLogCategoriesToEnable array = [
+  ''
```
AlexanderSehr, line 80: Not allLogs here?

ahmadabdalla: Not supported using the standard method. Was getting failures for it. New audit logging ingestion is being introduced in preview

modules/cache/redis-enterprise/.test/common/main.test.bicep
ahmadabdalla: Based on the spec, there's an additional object that can be inserted here:

modules: [
      {
        args: 'string'
        name: 'string'
      }
    ]

Was getting provisioning failures on the settings defined in the doco, without clear message why.

https://learn.microsoft.com/en-us/azure/templates/microsoft.cache/redisenterprise/databases?pivots=deployment-language-bicep

But don't think we should suppress it. Thoughts?

modules/cache/redis-enterprise/main.bicep
AlexanderSehr: Hm ok. Should it then even be an option? Maybe we can comment it out and add a disclaimer as to why

modules/cache/redis-enterprise/.test/common/main.test.bicep
AlexanderSehr: @ahmadabdalla If it's supported I guess we should allow it, even if we can't get it to work.

ahmadabdalla: I'm updating to add a parameter usage section too. Let me do some digging to try to get a working example.

ahmadabdalla: @AlexanderSehr finally got it to work :) At least now its tested

image

AlexanderSehr (approved):

PR #4022, opened 2023-09-25
[Utilities] Enabled `Set-ModuleReadMe` to find looped test cases

AlexanderSehr:
# Description

- Enabled `Set-ModuleReadMe` to find looped test cases
- Follow up to #4019

# Tests
- [x] Regenerated all docs to check that nothing changes
- [x] Tested loop case to check that nothing changes


AlexanderSehr:
Hey @tyconsulting, you may want to include this change in your environment too (in case you use usage examples)

ahmadabdalla (approved):

Issue #1210, opened 2022-04-03
SQL Managed Instances - Vulnerability Assessment with Storage Accounts using Private Endpoints


SenthuranSivananthan:
## Description

When using Storage Accounts with Private Endpoints, we have to use Managed Identity of SQL MI to turn on Vulnerability Assessment.  With MSI, the need for you no longer need to set `storageAccountAccessKey` value.  When this value is set, the deployment never completes.

### Steps to reproduce

1.  Create a storage account with private endpoints
2.  Create SQLMI with MSI
3.  RBAC SQLMI MSI -> Storage Account as Blob Contributor
4.  Enable Vulnerability Assessment

The deployment hangs.

Resolution:

1. Remove `storageAccountAccessKey` setting:  https://github.com/Azure/ResourceModules/blob/main/arm/Microsoft.Sql/managedInstances/vulnerabilityAssessments/deploy.bicep#L43

Ask:  Can this be configurable based on a parameter so access key is optionally used if PE is not enabled?

MrMCake:
Related of #1177

fblix:
waiting for #1964 to be completed.

PR #4033, opened 2023-09-27
[Modules] sql/managed-instance - Fixed bug with vulnerability assessment using storage acct. behind firewall/vnet

lsnoddy:
# Description

This PR fixes issue #1210.

 - Added a parameter to vulnerability-assessment (useStorageAccountAccessKey)  defaulted to false to indicate whether an access key should be used to access the storage account. If false, then the SQL MI managed indentity will be used.
 - Added parameter createStorageRoleAssignment  to indicate whether to assign the Storage Blob Data Contributor  role to the SQL Managed Instance system assigned identity to enable vulnerability assessment data to be written to the storage account as required. Used a .bicep directory under vulnerability-assessment with "nested_storageRoleAssignment.bicep" file to handle creating the role assignment due to the different scope.
 - storageAccountResourceId paramer is required (as per documentation)
 - vulnerabilty-assessmet is conditionally deployed based on systemAssignedIdentity parameter set to true

### Pipeline references

| Pipeline |
| - |
[Sql - ManagedInstances](https://github.com/lsnoddy/ResourceModules/actions/workflows/ms.sql.managedinstances.yml)
| |

## Type of Change

- [ ] Update to documentation

## Checklist

- [x] I did format my code


ahmadabdalla:
@lsnoddy pinged you offline for a sync up regarding this PR.

lsnoddy:
> @lsnoddy pinged you offline for a sync up regarding this PR.

After reviewing with @ahmadabdalla and Tao Yang, I think Tao's implementation using the "useStorageAccountAccessKey" is a more straightforward solution. I've updated the code to use that parameter instead. Also, any feedback on the Storage Blob Contributor role assignment implementation?

modules/sql/managed-instance/main.bicep
```diff
@@ -340,7 +340,7 @@ module managedInstance_vulnerabilityAssessment 'vulnerability-assessment/main.bi
     recurringScansEmailSubscriptionAdmins: contains(vulnerabilityAssessmentsObj, 'recurringScansEmailSubscriptionAdmins') ? vulnerabilityAssessmentsObj.recurringScansEmailSubscriptionAdmins : false
     recurringScansIsEnabled: contains(vulnerabilityAssessmentsObj, 'recurringScansIsEnabled') ? vulnerabilityAssessmentsObj.recurringScansIsEnabled : false
     storageAccountResourceId: vulnerabilityAssessmentsObj.storageAccountResourceId
-    storageAccountFirewallOrVnet: contains(vulnerabilityAssessmentsObj, 'storageAccountFirewallOrVnet') ? vulnerabilityAssessmentsObj.storageAccountFirewallOrVnet : false
```
AlexanderSehr: Hey @lsnoddy, how could come the other parameter is removed?

modules/sql/managed-instance/main.bicep
```diff
@@ -340,7 +340,7 @@ module managedInstance_vulnerabilityAssessment 'vulnerability-assessment/main.bi
...
+    useStorageAccountAccessKey: contains(vulnerabilityAssessmentsObj, 'useStorageAccountAccessKey') ? vulnerabilityAssessmentsObj.useStorageAccountAccessKey : true
```
AlexanderSehr, line 343: I wonder if this should be false. Not using an access key but other means of authentication seem more secure, right?

AlexanderSehr: Also PSRule seems to be after [AAD Auth](https://github.com/lsnoddy/ResourceModules/actions/runs/6331186963/attempts/1#summary-17195207525). It sounds related, but I'm not sure

modules/sql/managed-instance/vulnerability-assessment/main.bicep
```diff
@@ -39,12 +44,21 @@ resource managedInstance 'Microsoft.Sql/managedInstances@2022-02-01-preview' exi
   name: managedInstanceName
 }
 
+module roleAssignment '.bicep/nested_storageRoleAssignment.bicep' = if (!(useStorageAccountAccessKey)) {
+  name: guid(managedInstance.id, managedInstance.name, 'Storage Blob Data Contributor')
```
AlexanderSehr, line 48:
```suggestion
  name: guid(managedInstance.id, managedInstance.entity.principalId, 'Storage Blob Data Contributor')
```

modules/sql/managed-instance/vulnerability-assessment/main.bicep
```diff
@@ -39,12 +44,21 @@ resource managedInstance 'Microsoft.Sql/managedInstances@2022-02-01-preview' exi
...
+  scope: resourceGroup(splitStorageAccountResourceId[4])
+  params: {
+    storageAccountName: last(splitStorageAccountResourceId)
+    managedInstancePrincipalId: managedInstance.identity.principalId
```
AlexanderSehr, line 52: Just curious, is there a general requirement that it must be a system-assigned identity? And if it can also be a User-Assigned Identity, should we support it? Also, does this break omif one did not active the system assigned identity as it would then try and access a property which isn't available (if I'm not mistaken)

AlexanderSehr: This isn't easy to solve 😅

modules/sql/managed-instance/vulnerability-assessment/.bicep/nested_storageRoleAssignment.bicep
```diff
@@ -0,0 +1,19 @@
+@description('Required. The name of the storage account to apply the role assignment to.')
+param storageAccountName string
+
+@description('Required. The principal ID of the SQL managed instance system assigned managed identity to assign the role to.')
+param managedInstancePrincipalId string
+
+resource storageAccount 'Microsoft.Storage/storageAccounts@2021-09-01' existing = {
```
AlexanderSehr, line 7: Note, if we implement it like this, the storage account must be in the same RG as the SQL MI.

modules/sql/managed-instance/vulnerability-assessment/main.bicep
AlexanderSehr: We should also probably think about adding a test case for this.

ahmadabdalla: Scenarios potentially exist for RBAC:
1. SQL MI uses System Identity, and the existing Storage account is in the same RG.
2. SQL MI uses System Identity, and the existing Storage account is in a different RG.
3. SQL MI uses User Identity, and the existing Storage account is in the same RG.
4. SQL MI uses User Identity, and the existing Storage account is in a different RG.

3/4: I think it's better 'Not' to perform a Role assignment as this can be treated as a pre-req since the User identity would also exist before the SQL MI. 

1/2: I think we can optionally support this scenario, by an explicit parameter (i.e. `createStorageRoleAssignment`) to perform RBAC instead of doing it automatically if `useStorageAccountAccessKey` is empty. If it's automatic, we are assuming the user/pipeline creating SQL MI has `owner` rights on either the same RG or the different RG containing the storage, which could also assume its Sub level `owner` role as well thus failing the deployment for end users if they don't have the permissions. 

Thoughts?

lsnoddy: According to the documentation, user assigned managed identities  are not supported, so the requirement would be to use system assigned. https://learn.microsoft.com/en-us/azure/azure-sql/database/sql-database-vulnerability-assessment-storage?view=azuresql#prerequisites, so we'd be looking at just scenarios 1/2. 

Regarding RBAC, it's indeed a valid point that if it's automatic RBAC permissions would be required. Initially I was thinking  we make this assumption for all modules, however in this case since it would be automatic, the user is unable to choose not to use the functionality which could fail the deployment. If we are going to support this scenario, I agree that we'd probably need a parameter to perform RBAC as you suggested.

modules/sql/managed-instance/vulnerability-assessment/.bicep/nested_storageRoleAssignment.bicep
lsnoddy: This works becuase back in vulnerability-assessment main.bicep, the call to this module sets the scope based on the storage acct resource group, then it passes in stg acct name to allow us just to use existing. I tested on a stg acct in a different resource group and it worked without issue.

modules/sql/managed-instance/vulnerability-assessment/main.bicep
lsnoddy: Using the MI principal ID throws this error: This expression is being used in an assignment to the "name" property of the "module" type, which requires a value that can be calculated at the start of the deployment. Properties of managedInstance which can be calculated at the start include "apiVersion", "id", "name", "type"

modules/sql/managed-instance/main.bicep
lsnoddy: I agree here. It probably makes better sense from a security standpoint to defuault to false.

modules/sql/managed-instance/vulnerability-assessment/main.bicep
lsnoddy: Met with @AlexanderSehr and we agreed it makes sense to use the createStorageRoleAssignment param to indicate whether to create the role assignment. Even though this would fail if the role assignment already exists on the storage account, performing the role assignment more closely matches the behavior in the portal. I added a note on the parameter to caution on this.

modules/sql/managed-instance/.test/varbac/main.test.bicep
```diff
@@ -0,0 +1,87 @@
...
+
+@description('Optional. The name of the resource group to deploy for testing purposes.')
+@maxLength(90)
+param resourceGroupName string = 'ms.sql.managedinstances-${serviceShort}-rg'
+
+@description('Optional. The location to deploy resources to.')
+param location string = deployment().location
+
+@description('Optional. A short identifier for the kind of deployment. Should be kept short to not run into resource-name length-constraints.')
+param serviceShort string = 'sqlmimin'
```
AlexanderSehr: Won't that short not conflict with the `min` test?

modules/sql/managed-instance/.test/varbac/main.test.bicep
```diff
@@ -0,0 +1,87 @@
...
+
+@description('Optional. The password to leverage for the login.')
+@secure()
+param password string = newGuid()
+
+@description('Optional. Enable telemetry via a Globally Unique Identifier (GUID).')
+param enableDefaultTelemetry bool = true
+
+@description('Optional. A token to inject into the name of each resource.')
+param namePrefix string = 'lsno'
```
AlexanderSehr, line 25: Should be a token

modules/sql/managed-instance/README.md
```diff
@@ -592,3 +592,112 @@ module managedInstance './sql/managed-instance/main.bicep' = {
 
 </details>
 <p>
+
+<h3>Example 3: Varbac</h3>
```
AlexanderSehr, line 596: I'd recommend to call the folder just something like 'vulnAssm' or similar. That we run RBAC in the test is true, but I wonder if this should be the one thing we highlight :D

modules/sql/managed-instance/.bicep/nested_storageRoleAssignment.bicep
```diff
@@ -0,0 +1,17 @@
+param storageAccountName string
```
AlexanderSehr, line 1: I think your original location in the vuln. ass. folder was better.

modules/sql/managed-instance/.test/varbac/main.test.bicep
lsnoddy: Fixed the serviceShort string.

modules/sql/managed-instance/.test/varbac/main.test.bicep
lsnoddy: Left over from local testing, reverted back to token

modules/sql/managed-instance/README.md
lsnoddy: I like it, updated to 'vulnAssm'.

modules/sql/managed-instance/.bicep/nested_storageRoleAssignment.bicep
lsnoddy: Reverted back to original location.

settings.yml
```diff
@@ -12,8 +12,8 @@ variables:
   ########################
 
   # the 'localToken_' prefix will be removed from the key name when the pipelines run.
-  # e.g. if you have a token in your parameter file as [[customKey]], then the token defined in this file looks like "localToken_customKey": 'value'
-  localToken_namePrefix: '' # A 3-5 character length unique string, included in the resources names (e.g. 'cntso'). Used for local module testing and pipelines.
+  # e.g. if you have a token in your parameter file as <<customKey>>, then the token defined in this file looks like "localToken_customKey": 'value'
+  localToken_namePrefix: 'lsno' # A 3-5 character length unique string, included in the resources names (e.g. 'cntso'). Used for local module testing and pipelines.
```
AlexanderSehr, line 16: Another leftover 😉

AlexanderSehr: I guess it makes sense to undo all changes in this file

lsnoddy: Yes, needed my local settings.yml for pipeline run in my branch. Reverted to carml settings now that pipeline is green.

AlexanderSehr (approved):

PR #4084, opened 2023-10-14
Regenerated modules 1-60 

AlexanderSehr:
# Description

- Updated modules [60/366]

## Type of Change

- [ ] Update to documentation


modules/analysis-services/server/README.md
```diff
@@ -19,63 +19,30 @@ This module deploys an Analysis Services Server.
...
-## Outputs
+## Usage examples
 
-| Output Name | Type | Description |
-| :-- | :-- | :-- |
-| `location` | string | The location the resource was deployed into. |
-| `name` | string | The name of the analysis service. |
-| `resourceGroupName` | string | The resource group the analysis service was deployed into. |
-| `resourceId` | string | The resource ID of the analysis service. |
+The following module usage examples are retrieved from the content of the files hosted in the module's `tests` folder.
```
ahmadabdalla, line 24: Could this be reworded for example to not necessarily call out the test folder? What if a user has access to a readme, and the BR/TS, but not the source code? 

E.g. '_The following section provides usage examples for the module, by inputting various parameter combinations, which were then used to validate and deploy the module successfully.'_

Thoughts?

AlexanderSehr: @eriqua, I think Ahmad is on to something

AlexanderSehr: I wonder though if we shouldn't still point them somewhere as users would otherwise have no idea. We could for example link the "to be" public registry page

eriqua: Good points both! I agree we should somehow ref the module source code. The challenge is the version, which is not tracked by the readme, so at this stage we cannot be very specific, but just point to the source code repo in general. I'm fine with @AlexanderSehr suggestion. And thanks @ahmadabdalla for raising this!

AlexanderSehr: Updated in all PRs

ahmadabdalla: Yep all good @AlexanderSehr and @eriqua ✌️

PR #4124, opened 2023-10-21
[Modules] Updated identities to UDT as per AVM specs

krbar:
# Description

- Updated identities to UDT as per AVM specs
- Regenerated all JSON templates & readmes
- Small fixes here and there to account for failing test cases
- 
### Pipeline references

| Pipeline |
| - |
| |

## Type of Change

- [ ] Update to documentation

## Checklist

- [ ] I did format my code


modules/app-configuration/configuration-store/.test/common/main.test.bicep
```diff
@@ -103,9 +103,8 @@ module testDeployment '../../main.bicep' = {
       }
     ]
     softDeleteRetentionInDays: 1
-    systemAssignedIdentity: false
-    userAssignedIdentities: {
-      '${nestedDependencies.outputs.managedIdentityResourceId}': {}
+    managedIdentities: {
```
AlexanderSehr, line 106: Is this flip intended? Before it had User-assigned identities

krbar: Thanks, it was not intended, I switched systemAssigned back to false.

krbar: @AlexanderSehr just found out that the resource supports both identity types at the same time. I will switch systemAssigned to back to "true" if you don't mind

modules/app-configuration/configuration-store/main.bicep
```diff
@@ -82,12 +79,12 @@ param privateEndpoints privateEndpointType
...
 var enableReferencedModulesTelemetry = false
 
-var identityType = systemAssignedIdentity ? 'SystemAssigned' : !empty(userAssignedIdentities) ? 'UserAssigned' : 'None'
+var formattedUserAssignedIdentities = reduce(map((managedIdentities.?userAssignedResourcesIds ?? []), (id) => { '${id}': {} }), {}, (cur, next) => union(cur, next)) // Converts the flat array to an object like { '${id1}': {}, '${id2}': {} }
 
-var identity = {
-  type: identityType
-  userAssignedIdentities: !empty(userAssignedIdentities) ? userAssignedIdentities : null
-}
+var identity = !empty(managedIdentities) ? {
```
AlexanderSehr, line 84: As per original template this may need to be only one or the other

AlexanderSehr: Of so, we should probably mention this in the description

modules/cache/redis/main.bicep
```diff
@@ -110,12 +107,12 @@ param enableDefaultTelemetry bool = true
...
 
-var identityType = systemAssignedIdentity ? 'SystemAssigned' : !empty(userAssignedIdentities) ? 'UserAssigned' : 'None'
+var formattedUserAssignedIdentities = reduce(map((managedIdentities.?userAssignedResourcesIds ?? []), (id) => { '${id}': {} }), {}, (cur, next) => union(cur, next)) // Converts the flat array to an object like { '${id1}': {}, '${id2}': {} }
 
-var identity = {
-  type: identityType
-  userAssignedIdentities: !empty(userAssignedIdentities) ? userAssignedIdentities : null
-}
+var identity = !empty(managedIdentities) ? {
+  type: (managedIdentities.?systemAssigned ?? false) ? (!empty(managedIdentities.?userAssignedResourcesIds ?? {}) ? 'SystemAssigned,UserAssigned' : 'SystemAssigned') : (!empty(managedIdentities.?userAssignedResourcesIds ?? {}) ? 'UserAssigned' : null)
```
AlexanderSehr, line 113: As per original template this may need to be only one or the other. If so, we should probably mention this in the description.

modules/web/hosting-environment/main.bicep
```diff
@@ -154,7 +152,7 @@ resource appServiceEnvironment 'Microsoft.Web/hostingEnvironments@2022-03-01' =
   kind: kind
   location: location
   tags: tags
-  identity: identity
+  identity: identity!
```
AlexanderSehr, line 155: I guess that's a warning? But using `!` as a means to claim it will never be null seems wrong.

modules/signal-r-service/web-pub-sub/main.bicep
```diff
@@ -30,11 +30,8 @@ param capacity int = 1
 @description('Optional. Pricing tier of the resource.')
 param sku string = 'Standard_S1'
 
-@description('Optional. Enables system assigned managed identity on the resource.')
-param systemAssignedIdentity bool = false
-
-@description('Optional. The ID(s) to assign to the resource.')
-param userAssignedIdentities object = {}
+@description('Optional. The managed identity definition for this resource.')
```
AlexanderSehr, line 33: I wonder if we should mention that it can only be either or

modules/operational-insights/workspace/main.bicep
```diff
@@ -73,11 +73,8 @@ param publicNetworkAccessForIngestion string = 'Enabled'
 ])
 param publicNetworkAccessForQuery string = 'Enabled'
 
-@description('Optional. Enables system assigned managed identity on the resource.')
-param systemAssignedIdentity bool = false
-
-@description('Optional. The ID(s) to assign to the resource.')
-param userAssignedIdentities object = {}
+@description('Optional. The managed identity definition for this resource.')
```
AlexanderSehr, line 76: I wonder if we should mention here that it can only be either system assigned OR user-assigned

modules/machine-learning-services/workspace/main.bicep
```diff
@@ -56,12 +56,10 @@ param tags object = {}
...
 param enableDefaultTelemetry bool = true
 
-// Identity
-@sys.description('Conditional. Enables system assigned managed identity on the resource. Required if `userAssignedIdentities` is not provided.')
-param systemAssignedIdentity bool = false
-
-@sys.description('Conditional. The ID(s) to assign to the resource. Required if `systemAssignedIdentity` is set to false.')
-param userAssignedIdentities object = {}
+@sys.description('Optional. The managed identity definition for this resource. At least one identity type is required.')
+param managedIdentities managedIdentitiesType = {
```
AlexanderSehr, line 60: Why the default?

modules/document-db/database-account/gremlin-database/main.bicep
```diff
@@ -60,10 +62,7 @@ resource gremlinDatabase 'Microsoft.DocumentDB/databaseAccounts/gremlinDatabases
   name: name
   tags: tags
   parent: databaseAccount
-  identity: (identityType != 'None' ? {
-    type: identityType
-    userAssignedIdentities: !empty(userAssignedIdentities) ? userAssignedIdentities : null
-  } : null)!
+  identity: identity!
```
AlexanderSehr, line 65: Same as in the other case, is `!` a good idea here?

modules/compute/disk-encryption-set/main.bicep
```diff
@@ -91,7 +90,7 @@ resource keyVault 'Microsoft.KeyVault/vaults@2021-10-01' existing = {
 }
 
 // Note: This is only enabled for user-assigned identities as the service's system-assigned identity isn't available during its initial deployment
-module keyVaultPermissions 'modules/nested_keyVaultPermissions.bicep' = [for (userAssignedIdentityId, index) in items(userAssignedIdentities): {
+module keyVaultPermissions 'modules/nested_keyVaultPermissions.bicep' = [for (userAssignedIdentityId, index) in items(formattedUserAssignedIdentities): {
```
AlexanderSehr, line 93: Maybe it would be easier to directly reference the user assigned identities. For example via
`[for (userAssignedIdentityResourceId, index) in (managedIdentities.?userAssignedResourcesIds ?? []): {`

and update also the reference inside the loop to simply `userAssignedIdentityResourceId` instead of the `.key` reference.

modules/app-configuration/configuration-store/main.bicep
krbar: Contrary to the previous implementation, the resource supports any combination of identities ([see template reference](https://learn.microsoft.com/en-us/azure/templates/microsoft.appconfiguration/configurationstores?pivots=deployment-language-bicep#resourceidentity)), so I changed the implementation accordingly.

modules/cache/redis/main.bicep
krbar: Another case of a resource, which supports any combination of identities, even if the previous version of the module suggested that only one identity is supported ([template reference](https://learn.microsoft.com/en-us/azure/templates/microsoft.cache/redis?pivots=deployment-language-bicep#managedserviceidentity)). The test case "common" with both types of identities works fine.

modules/operational-insights/workspace/main.bicep
krbar: Added it to all modules like this (commit https://github.com/Azure/ResourceModules/pull/4124/commits/eb53c23b5848180c48399a826001d861f0588bfc). If you have suggestions regarding the wording, please let me know.

modules/signal-r-service/web-pub-sub/main.bicep
krbar: Added it to all modules like this (commit https://github.com/Azure/ResourceModules/pull/4124/commits/eb53c23b5848180c48399a826001d861f0588bfc). If you have suggestions regarding the wording, please let me know.

modules/web/hosting-environment/main.bicep
krbar: As discussed, changed it to any() in the definition of the "identity" variable, which also suppresses the warning.

modules/machine-learning-services/workspace/main.bicep
krbar: It's because at least one identity type is required for this resource type (the same for `disc-encryption-set`).

modules/compute/disk-encryption-set/main.bicep
krbar: Implemented as suggested: https://github.com/Azure/ResourceModules/pull/4124/commits/239d045c98ac99e4921240dde8f69e40077ea4a0

modules/document-db/database-account/gremlin-database/main.bicep
krbar: It looks like there is no identity support for the Gremlin Database, neither in the documentation nor in the module (this parameter is not passed on to the child resource) nor in the DB settings in the portal. This means that the warning is legit. I've removed the identity support from the child-resource. (commit: https://github.com/Azure/ResourceModules/pull/4124/commits/7dbf049f8b6217fc58fec4094e3c8e68d10b8c78)

AlexanderSehr (approved):

Issue #3386, opened 2023-06-29
[Bug Report]: resources/resource-group - ResourceGroupManagedByMismatch


coolhome:
### Describe the bug

Azure defaults `ResourceGroup.managedBy` as `null` but the ResourceModule template defaults to empty string. It is also not possible to set null with this module.

![image](https://github.com/Azure/ResourceModules/assets/700740/f7e080d8-9aec-4fd1-9575-b4732539497f)

If I check if empty and send null then I get another error:
```json
InvalidTemplate - Long running operation failed with status 'Failed'.
     | Additional Info:'Deployment template validation failed: 'The template
     | resource 'RG-Workday' at line '1' and column '2529' is not valid:
     | Evaluation result of language expression
     | '[if(empty(parameters('managedBy')), null(), parameters('managedBy'))]'
     | is type 'Null', expected type is 'String'.. Please see
     | https://aka.ms/arm-functions for usage details.'.
```

### To reproduce

1. Create a resource group through azure portal
2. Create bicep for this resource group
```bicep 
module ResourceGroup '.carml/resources/resource-groups/main.bicep' = {
  name: '${deployment().name}-ResourceGroup'
  params: {
    name: resourceGroupName
    location: resourceGroupLocation
  }
}
```
3. Deploy Bicep
```
The managed by property of the resource group cannot be changed from its current value '<null>'
```

### Relevant log output

```Shell
{
  "code": "DeploymentFailed",
  "target": "/subscriptions/***",
  "message": "At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details.",
  "details": [
    {
      "code": "ResourceGroupManagedByMismatch",
      "message": "The managed by property of the resource group cannot be changed from its current value '<null>'."
    }
  ]
}
```

zydio:
This also happens to any Resource Group deployed using CARML modules with v <= 0.8.0, when trying to deploy again the same BICEP with CARML v >= 0.9.0 when this change has been introduced with PR https://github.com/Azure/ResourceModules/pull/2319 which was also erroneously marked as non-breaking change.. but as stated this breaks any existing deployment.

adnkai:
Any news on this?

coolhome:
I wonder if we should report this as a Resource Provider bug because I think the value should always be an immutable string. Alternatively, maybe there should be a ManagedResourceGroup module that separates this functionality out. 

If you consume this module, I suggest removing the managedBy param.

[...]

PR #4351, opened 2023-12-07
Resource Group - ManagedBy NULL/Empty String issue

coolhome:
# Description

Potential fix for #3386

This allowed me to omit the value. **I have not tested lock module or the role assignment resource.** 

Unfortunately, this is a breaking change as `managedBy` is implicitly null and cannot be defaulted to an empty string. However, I think this may be desired since Azure CLI/PowerShell set this as null.

Please feel free to pick up and continue with testing.

### Pipeline references

| Pipeline |
| - |
| |

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Update to documentation

## Checklist

- [X] I did format my code

### DIFF
--- a/modules/resources/resource-group/main.bicep
+++ b/modules/resources/resource-group/main.bicep
@@ -20,7 +20,7 @@ param roleAssignments roleAssignmentType
 param tags object?
 
 @description('Optional. The ID of the resource that manages this resource group.')
-param managedBy string = ''
+param managedBy string?
 
 @description('Optional. Enable telemetry via a Globally Unique Identifier (GUID).')
 param enableDefaultTelemetry bool = true
@@ -51,11 +51,18 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
   }
 }
 
-resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
+resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = if (managedBy == null) {
   location: location
   name: name
   tags: tags
-  managedBy: managedBy
+  properties: {}
+}
+
+resource managedResourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = if (managedBy != null) {
+  location: location
+  name: name
+  tags: tags
+  managedBy: length(managedBy ?? '') > 0 ? managedBy : ''
   properties: {}
 }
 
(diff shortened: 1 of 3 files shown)


AlexanderSehr:
Hey @coolhome, thanks for the PR :) Sorry that I didn't get around to it earlier - yet alone the linked issue. As the module is already migrated to AVM, @segraef, could you take a look at this? As @coolhome says, we should also test this out, but if it checks out, it would be great to get updated in the [resource group](https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/resources/resource-group) module.

modules/resources/resource-group/main.bicep
```diff
@@ -51,25 +51,32 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
...
   tags: tags
-  managedBy: managedBy
+  properties: {}
+}
+
+resource managedResourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = if (managedBy != null) {
+  location: location
+  name: name
+  tags: tags
+  managedBy: length(managedBy ?? '') > 0 ? managedBy : ''
```
AlexanderSehr, line 65:
```suggestion
  managedBy: managedBy ?? ''
```
If I'm not mistaken, this should do the trick to. Even better would be only `mangedBy: mangedBy`, IF `null` is accepted by the RP.

segraef:
> Hey @coolhome, thanks for the PR :) Sorry that I didn't get around to it earlier - yet alone the linked issue. As the module is already migrated to AVM, @segraef, could you take a look at this? As @coolhome says, we should also test this out, but if it checks out, it would be great to get updated in the [resource group](https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/resources/resource-group) module.

Testing ...

modules/resources/resource-group/main.bicep
coolhome: That is correct, but the RP does not accept `null`. If its json property exists it has to be a string, which makes this problem hard. This is why I've had challenges figuring out who to report this to CARML/AVM, Bicep, or Azure Support for RP.

I think the RP should accept null and require a string length greater than 0. Regardless it's a breaking change if someone used this module to create the resource group because it has to be deleted to remove managedBy AFAIK.

AlexanderSehr: In the corresponding AVM module the owner now opted to not expose the ability to write to this property. This won't impact the RP but at least it wouldn't cause any other issues for users of the module.

PR #4388, opened 2023-12-19
Fix `winrm` listeners data type

coolhome:
# Description

winRM is passed to listeners, which is an array and not an object. I was not able to perform set-module on the virtual machine construct, that template is broken.

### Pipeline references

| Pipeline |
| - |
| |

## Type of Change

- [X] Update to documentation

## Checklist

- [X] I did format my code

### DIFF
--- a/constructs/Compute/virtualMachinesMultiple/main.bicep
+++ b/constructs/Compute/virtualMachinesMultiple/main.bicep
@@ -285,7 +285,7 @@ param timeZone string = ''
 param additionalUnattendContent array = []
 
 @description('Optional. Specifies the Windows Remote Management listeners. This enables remote Windows PowerShell. - WinRMConfiguration object.')
-param winRM object = {}
+param winRM array = {}
 
 @description('Optional. Any VM configuration profile assignments.')
 param configurationProfileAssignments string = ''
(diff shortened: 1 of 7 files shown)


AlexanderSehr:
Hey @coolhome,
the construct template definitely is. It's essentially just code we parked when initially migrating all modules from ARM to Bicep. In the future setup with [AVM](https://aka.ms/avm), it would be considered a pattern.
I think the reason we did not outright delete it was that it contains logic to generate x amount of VM names for the deploying person.

constructs/Compute/virtualMachinesMultiple/main.bicep
```diff
@@ -285,7 +285,7 @@ param timeZone string = ''
 param additionalUnattendContent array = []
 
 @description('Optional. Specifies the Windows Remote Management listeners. This enables remote Windows PowerShell. - WinRMConfiguration object.')
```
coolhome, line 287: I do want to mention that this looks like it meant to be a type [WinRMConfiguration](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?view=rest-compute-2023-07-01) but in actuality its [WinRMListener](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?view=rest-compute-2023-07-01). Maybe the solution is to remove the listener mapping and allow users to pass the WinRMConfiguration type?

AlexanderSehr:
@rahalan, note this update may be relevant for the [AVM VM PR](https://github.com/Azure/bicep-registry-modules/pull/717). However, @coolhome could it be the encoding or something similar was changed? The files appears completely new which makes it very hard to sport what actually changed.

coolhome:
@AlexanderSehr yah I just noticed that as well. I have `core.autocrlf=false` and I think this is the root cause.

image
This is not really a solution but a way to see the differences without the newline issue. 

Let me know if you want to re-create the PR. I am also happy with whatever.

coolhome:
Closing since its now implemented in https://github.com/Azure/bicep-registry-modules/pull/717 for AVM.

AlexanderSehr:
Hey @coolhome, may I ask you to re-open the PR? I think for VMSS this is still very much relevant, isn't it? This module isn't migrated yet / started.

TheMarcosP:
h

AlexanderSehr:
Hey @rahlan, I take it it's save to say you don't have this on your radar these days.
Could you please check if the contribution is already implemented in the VM/VMSS module and if not ideally consider to add it?

Issue #4035, opened 2023-09-28
[Feature Request]: Bump API version and add additional parameters for Insights.PrivateLinkScopes


akata72:
### Description

Please bump the API version (->2021-07-01-preview) of the microsoft.insights/privatelinkscopes and add support for the new properties. 

 ```
properties: {
    accessModeSettings: {
      exclusions: []
      queryAccessMode: 'Open'
      ingestionAccessMode: 'Open'
    }
  }
```

johnlokerse:
@AlexanderSher Please assign me to this issue. I need this in my solution too. 😄

johnlokerse:
@AlexanderSehr Please assign me to this issue. I need this in my solution too. 😄

AlexanderSehr:
Resolved through linked PR

PR #4392, opened 2023-12-21
[Modules] Updates to the Azure Monitor Private Link Service (AMPLS) Module

johnlokerse:
# Description

Requested upgrade for AMPLS in issue #4035.

Upgraded the API version of the AMPLS resource and added the required properties. Also, added two UDTs to make parameters input for the user easier. This is a breaking change because it introduced new properties that are required in the newer version.

## Type of Change

- [ ] Update to documentation

## Checklist

- [x] I did format my code

### DIFF
--- a/modules/insights/private-link-scope/main.bicep
+++ b/modules/insights/private-link-scope/main.bicep
@@ -7,7 +7,7 @@ metadata owner = 'Azure/module-maintainers'
 param name string
 
 @description('''
-  Required. Specifies the access mode of ingestion or queries through associated private endpoints in scope. For security reasons, it is recommended to use PrivateOnly whenever possible to avoid data exfiltration.
+  Optional. Specifies the access mode of ingestion or queries through associated private endpoints in scope. For security reasons, it is recommended to use PrivateOnly whenever possible to avoid data exfiltration.
 
   * Private Only (default) - Allows the connected virtual network to reach only Private Link resources. This mode is the most secure.
   * Open - Allows the connected virtual network to reach both Private Link resources and the resources not in the AMPLS resource. Data exfiltration cannot be prevented in this mode.
--- a/modules/insights/private-link-scope/main.bicep
+++ b/modules/insights/private-link-scope/main.bicep
@@ -250,7 +250,7 @@ type scopedResourceType = {
 }[]?
 
 type accessModeType = {
-  @description('Optional. List of exclusions that override the default access mode settings for specific private endpoint connections.')
+  @description('Optional. List of exclusions that override the default access mode settings for specific private endpoint connections. Exclusions for the current created Private endpoints can only be applied post initial provisioning.')
   exclusions: {
     @description('Required. The private endpoint connection name associated to the private endpoint on which we want to apply the specific access mode settings.')
     privateEndpointConnectionName: string
--- a/modules/insights/private-link-scope/main.bicep
+++ b/modules/insights/private-link-scope/main.bicep
@@ -6,8 +6,16 @@ metadata owner = 'Azure/module-maintainers'
 @minLength(1)
 param name string
 
-@description('Required. Specifies the access mode of ingestion or queries through associated private endpoints in scope.')
-param accessModeSettings accessModeType
+@description('''
+  Required. Specifies the access mode of ingestion or queries through associated private endpoints in scope. For security reasons, it is recommended to use PrivateOnly whenever possible to avoid data exfiltration.
+
+  * Private Only (default) - Allows the connected virtual network to reach only Private Link resources. This mode is the most secure.
+  * Open - Allows the connected virtual network to reach both Private Link resources and the resources not in the AMPLS resource. Data exfiltration cannot be prevented in this mode.
+''')
+param accessModeSettings accessModeType = {
+  ingestionAccessMode: 'PrivateOnly'
+  queryAccessMode: 'PrivateOnly'
+}
 
 @description('Optional. The location of the private link scope. Should be global.')
 param location string = 'global'
--- a/modules/insights/private-link-scope/main.bicep
+++ b/modules/insights/private-link-scope/main.bicep
@@ -6,6 +6,9 @@ metadata owner = 'Azure/module-maintainers'
 @minLength(1)
 param name string
 
+@description('Required. Specifies the access mode of ingestion or queries through associated private endpoints in scope.')
+param accessModeSettings accessModeType
+
 @description('Optional. The location of the private link scope. Should be global.')
 param location string = 'global'
 
@@ -16,7 +19,7 @@ param lock lockType
 param roleAssignments roleAssignmentType
 
 @description('Optional. Configuration details for Azure Monitor Resources.')
-param scopedResources array = []
+param scopedResources scopedResourceType
 
 @description('Optional. Configuration details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible.')
 param privateEndpoints privateEndpointType
@@ -49,14 +52,16 @@ resource defaultTelemetry 'Microsoft.Resources/deployments@2021-04-01' = if (ena
   }
 }
 
-resource privateLinkScope 'Microsoft.Insights/privateLinkScopes@2019-10-17-preview' = {
+resource privateLinkScope 'microsoft.insights/privateLinkScopes@2021-07-01-preview' = {
   name: name
   location: location
   tags: tags
-  properties: {}
+  properties: {
+    accessModeSettings: accessModeSettings
+  }
 }
 
-module privateLinkScope_scopedResource 'scoped-resource/main.bicep' = [for (scopedResource, index) in scopedResources: {
+module privateLinkScope_scopedResource 'scoped-resource/main.bicep' = [for (scopedResource, index) in (scopedResources ?? []): {
   name: '${uniqueString(deployment().name, location)}-PvtLinkScope-ScopedRes-${index}'
   params: {
     name: scopedResource.name
@@ -227,3 +232,31 @@ type privateEndpointType = {
   @description('Optional. Enable/Disable usage telemetry for module.')
   enableTelemetry: bool?
 }[]?
+
+type scopedResourceType = {
+  @description('Required. Name of the private link scoped resource.')
+  name: string
+
+  @description('Required. The resource ID of the scoped Azure monitor resource.')
+  linkedResourceId: string
+}[]?
+
+type accessModeType = {
+  @description('Optional. List of exclusions that override the default access mode settings for specific private endpoint connections.')
+  exclusions: {
+    @description('Required. The private endpoint connection name associated to the private endpoint on which we want to apply the specific access mode settings.')
+    privateEndpointConnectionName: string
+
+    @description('Required. Specifies the access mode of ingestion through the specified private endpoint connection in the exclusion.')
+    ingestionAccessMode: 'Open' | 'PrivateOnly'
+
+    @description('Required. Specifies the access mode of queries through the specified private endpoint connection in the exclusion.')
+    queryAccessMode: 'Open' | 'PrivateOnly'
+  }[]?
+
+  @description('Required. Specifies the default access mode of ingestion through associated private endpoints in scope.')
+  ingestionAccessMode: 'Open' | 'PrivateOnly'
+
+  @description('Required. Specifies the default access mode of queries through associated private endpoints in scope.')
+  queryAccessMode: 'Open' | 'PrivateOnly'
+}
--- a/modules/insights/private-link-scope/main.bicep
+++ b/modules/insights/private-link-scope/main.bicep
@@ -12,10 +12,7 @@ param name string
   * Private Only (default) - Allows the connected virtual network to reach only Private Link resources. This mode is the most secure.
   * Open - Allows the connected virtual network to reach both Private Link resources and the resources not in the AMPLS resource. Data exfiltration cannot be prevented in this mode.
 ''')
-param accessModeSettings accessModeType = {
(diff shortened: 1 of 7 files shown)


modules/insights/private-link-scope/tests/e2e/defaults/main.test.bicep
```diff
@@ -45,5 +45,9 @@ module testDeployment '../../../main.bicep' = [for iteration in [ 'init', 'idem'
   params: {
     enableDefaultTelemetry: enableDefaultTelemetry
     name: '${namePrefix}${serviceShort}001'
+    accessModeSettings: {
```
AlexanderSehr, line 48: Hey @johnlokerse, 
A great contribution as always. One question though: Would you think it makes sense to use this value for `accessModeSettings` as a default - i.e. making this parameter optional, or should it rather be mandatory as it's very different from case-to-case.

modules/insights/private-link-scope/main.bicep
```diff
@@ -6,6 +6,9 @@ metadata owner = 'Azure/module-maintainers'
 @minLength(1)
 param name string
 
+@description('Required. Specifies the access mode of ingestion or queries through associated private endpoints in scope.')
```
AlexanderSehr, line 9: Please run the `utilities/tools/Set-Module` utility on the module folder to have it update both the `main.json` & `readme.md` files

johnlokerse:
@AlexanderSehr 

I have an issue with the "max" test. The latest API version introduces the `accessModeSettings` property which holds 3 properties: `ingestionAccessMode`, `queryAccessMode` and `exclusions` (optional).

The `exclusions` property is vague for me on how it works. In the Azure Portal you can choose the specific endpoint using a dropdown menu:

image

But, in the Bicep IaC it is a bit different because you can set any private endpoint connection name you want. It does not have to exist either and it holds any string value...

Also, what is weird to me is that the connection name is randomly generated: 
image

Because of this randomly generated connection name I am not able to potentially set the exclusion for this private endpoint, because the exclusion is set before the creation of the connection. This might be an issue on the resource side. What do you think?

modules/insights/private-link-scope/main.bicep
AlexanderSehr, line 1: Would you happen to have a pipeline badge / any proof of the tests passing handy? If not, we'll need an intermediate branch so that I can test the change on our end before merging to main.

johnlokerse: I will run the test in the pipeline when I have resolved the PR comments 😃

modules/insights/private-link-scope/tests/e2e/defaults/main.test.bicep
johnlokerse: We can set the default value for the required parameters. I think the default value is 'Open', but have to double-check this to make sure.

AlexanderSehr:
> @AlexanderSehr
> 
> I have an issue with the "max" test. The latest API version introduces the `accessModeSettings` property which holds 3 properties: `ingestionAccessMode`, `queryAccessMode` and `exclusions` (optional).
> 
> The `exclusions` property is vague for me on how it works. In the Azure Portal you can choose the specific endpoint using a dropdown menu:
> 
> image
> But, in the Bicep IaC it is a bit different because you can set any private endpoint connection name you want. It does not have to exist either and it holds any string value... See "thisisatest" for example.
> 
> Also, what is weird to me is that the connection name is randomly generated: image
> 
> Because of this randomly generated connection name I am not able to potentially set the exclusion for this private endpoint, because the exclusion is set before the creation of the connection. This might be an issue on the resource side. What do you think?

Great question @johnlokerse, let me look into this a bit more. While doing so I'll also loop in the to-be module owner in AVM, @ahmadabdalla as he may have had experience with this property.

modules/insights/private-link-scope/tests/e2e/defaults/main.test.bicep
AlexanderSehr: It is the default. Based on this statement one could however also argue the opposite would be the better option: 
image

Personally, I don't know the resource well enough to make an informed statement. It's always a balancing act in between security & usability.

ahmadabdalla: I would vote to have the `ingestionAccessMode` and `queryAccessMode` become optional and follow the ARM reference [documentation](https://learn.microsoft.com/en-us/azure/templates/microsoft.insights/privatelinkscopes?pivots=deployment-language-bicep#accessmodesettings). We could provide additional context into the description field like we do with other params around security recommendations. My preference would be to set them to `privateOnly` them if `privateEndpoints` are enabled like we do for other resources, but not sure if that's achievable in UDT.

johnlokerse: @AlexanderSehr @ahmadabdalla Happy new year! I was on holiday leave and going to pick this up again. I agree to make it optional and give it the default value `PrivateOnly`.

> My preference would be to set them to privateOnly them if privateEndpoints are enabled like we do for other resources, but not sure if that's achievable in UDT.

Wil check if this is possible.

johnlokerse: @ahmadabdalla It does not work to set the access mode to `PrivateOnly` via the UDT, but it can be achieved this way:

```bicep
@description('''
  Required. Specifies the access mode of ingestion or queries through associated private endpoints in scope. For security reasons, it is recommended to use PrivateOnly whenever possible to avoid data exfiltration.

  * Private Only (default) - Allows the connected virtual network to reach only Private Link resources. This mode is the most secure.
  * Open - Allows the connected virtual network to reach both Private Link resources and the resources not in the AMPLS resource. Data exfiltration cannot be prevented in this mode.
''')
param accessModeSettings accessModeType = {
  ingestionAccessMode: empty(privateEndpoints) ? 'Open' : 'PrivateOnly'
  queryAccessMode: empty(privateEndpoints) ? 'Open' : 'PrivateOnly'
}

@description('Optional. Configuration details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible.')
param privateEndpoints privateEndpointType
```

I have tested it in both scenario's:
* With an empty `privateEndpoints` parameter it sets access mode to 'Open'
* With a filled `privateEndpoints` parameter it sets access mode to 'PrivateOnly'

This way the accessModeSettings parameter is optional, but it adds a layer of complexity. I was thinking since we set the access mode programmatically maybe we should set is as a variable instead of a parameter.

What do you guys think @AlexanderSehr @ahmadabdalla ?

AlexanderSehr: Hey @johnlokerse, the issue with 'only'  a variable is that the user wouldn't have a way to overwrite the default behavior. In that sense I think it would always be good to have at least the option. That said, the logic does not have to be part of the parameter but could also be implemented where it is used. E.g.

```bicep
param accessModeSettings accessModeType

resource privateLinkScope 'microsoft.insights/privateLinkScopes@2021-07-01-preview' = {
  name: name
  location: location
  tags: tags
  properties: {
    accessModeSettings: accessModeSettings ?? {
      ingestionAccessMode: empty(privateEndpoints) ? 'Open' : 'PrivateOnly'
      queryAccessMode: empty(privateEndpoints) ? 'Open' : 'PrivateOnly'
    }
  }
}
```

Whether that's 'better' is debatable. The only point I'm trying to make is that we should keep the parameter as an option no matter what.

johnlokerse: Ah, yes I agree. Let's go for the option you're showing.

johnlokerse: Changes are pushed.

modules/insights/private-link-scope/main.bicep
johnlokerse: Done, will update when other changes are required.

johnlokerse:
> > @AlexanderSehr
> > I have an issue with the "max" test. The latest API version introduces the `accessModeSettings` property which holds 3 properties: `ingestionAccessMode`, `queryAccessMode` and `exclusions` (optional).
> > The `exclusions` property is vague for me on how it works. In the Azure Portal you can choose the specific endpoint using a dropdown menu:
> > image
> > But, in the Bicep IaC it is a bit different because you can set any private endpoint connection name you want. It does not have to exist either and it holds any string value... See "thisisatest" for example.
> > Also, what is weird to me is that the connection name is randomly generated: image
> > Because of this randomly generated connection name I am not able to potentially set the exclusion for this private endpoint, because the exclusion is set before the creation of the connection. This might be an issue on the resource side. What do you think?
> 
> Great question @johnlokerse, let me look into this a bit more. While doing so I'll also loop in the to-be module owner in AVM, @ahmadabdalla as he may have had experience with this property.

@ahmadabdalla Do you have an idea how this works? I have looked into this and I think it's not possible to get the private endpoint connection name (`connection-`) dynamically or set the name manually. I can get the tests running green with any value.

modules/insights/private-link-scope/main.bicep
```diff
@@ -6,6 +6,14 @@ metadata owner = 'Azure/module-maintainers'
 @minLength(1)
 param name string
 
+@description('''
+  Required. Specifies the access mode of ingestion or queries through associated private endpoints in scope. For security reasons, it is recommended to use PrivateOnly whenever possible to avoid data exfiltration.
```
ahmadabdalla, line 10: Shouldn't this be set to Optional?

ahmadabdalla: Never mind this comment. I just validated that latest API sets this as a required parameter

ahmadabdalla:
> @ahmadabdalla Do you have an idea how this works? I have looked into this and I think it's not possible to get the private endpoint connection name (`connection-`) dynamically or set the name manually. I can get the tests running green with any value.

@AlexanderSehr @johnlokerse After further investigation from my end, it appears that this is a chicken and egg situation if we are to configure this from the start via IaC. The `privateEndpointConnections` child resource type under the `insights/privateLinkScopes` fails when being setup independently via IaC. This is the resource that gets a random name. And if we try to export an existing template with this child resource being in there, the template fails deployment for this resource type, but still goes in and creates another one with a random name too.

Given that we can still inject a name via the test file and that name shows up in the Portal

image

image

Is to me at least is an acceptable outcome on the things we can control, because it seems something we cannot easily deal with unless we start introducing complex logic.

We can however provide context in the exclude description field to become more informative for the end user. E.g. _"Exclusions for the current created Private endpoints can only be applied post initial provisioning"_

modules/insights/private-link-scope/main.bicep
johnlokerse: Yes, correct. Fixed!

johnlokerse:
> > @ahmadabdalla Do you have an idea how this works? I have looked into this and I think it's not possible to get the private endpoint connection name (`connection-`) dynamically or set the name manually. I can get the tests running green with any value.
> 
> @AlexanderSehr @johnlokerse After further investigation from my end, it appears that this is a chicken and egg situation if we are to configure this from the start via IaC. The `privateEndpointConnections` child resource type under the `insights/privateLinkScopes` fails when being setup independently via IaC (at least in Bicep). This is the resource that gets a random name. And if we try to export an existing template with this child resource being in there, the template fails deployment for this resource type, but still goes in and creates another one with a random name too.
> 
> Given that we can still inject a name via the test file and that name shows up in the Portal
> 
> image
> 
> image
> 
> Is to me at least is an acceptable outcome on the things we can control, because it seems something we cannot easily deal with unless we start introducing complex logic.
> 
> We can however provide context in the exclude description field to become more informative for the end user. E.g. _"Exclusions for the current created Private endpoints can only be applied post initial provisioning"_

I agree. I have added a description to the `exclusion` property. 

@AlexanderSehr Going to run the required actions now 😄

johnlokerse:
[Insights - PrivateLinkScopes](https://github.com/johnlokerse/ResourceModules/actions/workflows/ms.insights.privatelinkscopes.yml)

🚀

modules/insights/private-link-scope/main.bicep
```diff
@@ -6,6 +6,12 @@ metadata owner = 'Azure/module-maintainers'
 @minLength(1)
 param name string
 
+@description('''Optional. Specifies the access mode of ingestion or queries through associated private endpoints in scope. For security reasons, it is recommended to use PrivateOnly whenever possible to avoid data exfiltration.
+
+  * Private Only (default) - Allows the connected virtual network to reach only Private Link resources. This mode is the most secure.
```
ahmadabdalla, line 11: Just one last comment from my end @johnlokerse.. the default is private only if private endpoints are configured. Can you please update the description to reflect the behaviour. For example if we deploy the "default" test case, the behaviour will be 'open'. Other than that I am good.. thank you so much for your contribution <3

ahmadabdalla (changes requested):

modules/insights/private-link-scope/main.bicep
johnlokerse: Also rerunning static tests just to be sure...

ahmadabdalla (approved):

AlexanderSehr:
Off we go. Thanks @johnlokerse & @ahmadabdalla . What a beauty of a service 👀

johnlokerse:
> One small recommendation (not relevant for the PR) for you @johnlokerse: Bicep is now at version [0.24.24](https://github.com/Azure/bicep/releases/tag/v0.24.24). I recommend a `choco upgrade bicep` ;)
> 
> image

Done, thanks! 😄

PR #4499, opened 2024-02-18
[Utilities] Enabled Workflow-Trigger function to trigger only for module diff (ported from AVM)

krbar:
Enabled Workflow-Trigger function to trigger only for module diff (functionality ported from AVM)

Example:

```
VERBOSE: Fetching current GitHub workflows
VERBOSE: Requested HTTP/1.1 GET with 0-byte payload
VERBOSE: Received HTTP/1.1 response of content type application/json of unknown size
VERBOSE: Content encoding: utf-8
VERBOSE: Requested HTTP/1.1 GET with 0-byte payload
VERBOSE: Received HTTP/1.1 response of content type application/json of unknown size
VERBOSE: Content encoding: utf-8
VERBOSE: Fetched [147] workflows
WARNING: Resource type [azuresecuritycenter] does not exist in the API / is custom. Falling back to it as default.
WARNING: Resource type [azuresecuritycenter] does not exist in the API / is custom. Falling back to it as default.
VERBOSE: As per 'diff', filtered workflows down to [42]
VERBOSE: Triggering GitHub workflows for branch [users/krbar/updateNamePrefix]
VERBOSE: Performing the operation "Invoke" on target "GitHub workflow [ms.aad.domainservices.yml] for branch [users/krbar/updateNamePrefix]".
```

### DIFF
--- a/modules/network/ip-group/main.bicep
+++ b/modules/network/ip-group/main.bicep
@@ -9,7 +9,7 @@ param name string
 @description('Optional. Location for all resources.')
 param location string = resourceGroup().location
 
-@description('Optional. IpAddresses/IpAddressPrefixes in the IpGroups resource.')
+@description('Optional. IpAddresses/IpAddressPrefixes in the IpGroups resource...')
 param ipAddresses array = []
 
 @description('Optional. The lock settings of the service.')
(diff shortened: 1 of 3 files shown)


AlexanderSehr (approved):

AlexanderSehr:
🚀

PR #4508, opened 2024-02-26
[Fixes] Added missing slash to module path split

AlexanderSehr:
# Description

- Based on found issue by a colleague where the script failed if 'modules' also appeared in other parts of the path
- The `/modules/` instead of `modules/` is in alignment with all other places where we split
- Tested & validated

## Type of Change

- [ ] Update to documentation


ChrisSidebotham (approved):
lgtm

